PR #296, opened 2026-02-20
fix: read-only properties emit { get; } instead of { get; set; }

### LINKED ISSUE #290: Read-only properties should emit { get; } not { get; set; }
## Problem
Properties defined with only `§GET` (no `§SET`) emit as `{ get; set; }` instead of `{ get; }`.

## Context
Discovered during C# → Calor conversion campaign. Affects multiple projects.

## Expected Behavior
```csharp
// Current (incorrect):
public string Number { get; set; }

// Expected (when only §GET is defined):
public string Number { get; }
```

## Files to Change
- `src/Calor.Compiler/CodeGen/CSharpEmitter.cs`

## Testing
- Add regression test
- Run `dotnet test -c Release` and `calor_self_test`

juanmicrosoft:
## Summary
- Fixes #290
- Auto-properties with only `§GET` (no `§SET`/`§INIT`) now emit `{ get; }` instead of `{ get; set; }`
- Properties with `§INIT` emit `{ get; init; }`
- Updated golden files for `09_codegen_bugfixes` scenario

## Test plan
- [x] Added 2 regression tests in `ConversionCampaignFixTests.cs`
- [x] All 3,471 tests pass
- [x] All 10/10 self-test scenarios pass
- [x] Golden files updated

### DIFF
--- a/src/Calor.Compiler/CodeGen/CSharpEmitter.cs
+++ b/src/Calor.Compiler/CodeGen/CSharpEmitter.cs
@@ -2332,14 +2332,17 @@ public sealed class CSharpEmitter : IAstVisitor<string>
         // Auto-property with default value
         if (node.IsAutoProperty)
         {
+            var accessors = node.Setter != null ? "get; set;" :
+                            node.Initer != null ? "get; init;" :
+                            "get;";
             if (node.DefaultValue != null)
             {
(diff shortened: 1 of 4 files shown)


juanmicrosoft:

PR #329, opened 2026-02-20
fix: strip @ prefix from C# verbatim identifiers

### LINKED ISSUE #309: Handle @-prefixed C# parameter names in converter
## Problem
C# @-prefixed identifiers (like @event, @class) are not stripped of the @ prefix during conversion.

## Files
- src/Calor.Compiler/Migration/RoslynSyntaxVisitor.cs

juanmicrosoft:
## Summary
- Uses `Identifier.ValueText` instead of `Identifier.Text` for parameters, variables, fields, and references
- Strips the `@` prefix that C# uses for keyword-as-identifier (e.g., `@class` → `class`, `@event` → `event`)

Closes #309

## Test plan
- [x] New regression tests in ConversionCampaignFixTests.cs (2 tests)
- [x] `dotnet test -c Release` passes (3,471 passed)


juanmicrosoft:

PR #336, opened 2026-02-20
fix: decimal type bind attribute parsing (#293)

### LINKED ISSUE #293: Fix decimal type variable/type emission order
## Problem
The emitter has issues with decimal type emission, producing incorrect variable declaration ordering or type name.

## Context
Discovered in Project #3 (NumbersInCSharp) during conversion campaign.

## Files to Change
- `src/Calor.Compiler/CodeGen/CSharpEmitter.cs`

## Testing
- Add regression test
- Run `dotnet test -c Release` and `calor_self_test`

juanmicrosoft:
## Summary
- Adds `dec`, `decimal`, and `double` to `IsLikelyType` in `AttributeHelper.cs`
- Fixes `§B{dec:x}` being parsed as name=dec, type=x (swapped) instead of type=decimal, name=x
- This caused C# emission like `x dec = 3.14m` instead of `decimal x = 3.14m`

## Root cause
`IsLikelyType` had `f32`/`f64`/`float` but was missing `dec`/`decimal`/`double`, so the bind attribute format detector couldn't distinguish `{type:name}` from `{name:type}` for these types.

## Test plan
- [x] New test verifying `§B{dec:x} 3.14m` emits `decimal x = 3.14m`
- [x] All 4,406 tests pass (0 failures)
- [x] All 10 self-test golden file scenarios pass

Fixes #293

### DIFF
--- a/src/Calor.Compiler/Parsing/AttributeHelper.cs
+++ b/src/Calor.Compiler/Parsing/AttributeHelper.cs
@@ -169,8 +169,8 @@ public static class AttributeHelper
         {
             "i8" or "i16" or "i32" or "i64" => true,
             "u8" or "u16" or "u32" or "u64" => true,
-            "f32" or "f64" => true,
-            "int" or "float" or "str" or "string" or "bool" or "void" or "char" or "var" => true,
+            "f32" or "f64" or "dec" or "decimal" => true,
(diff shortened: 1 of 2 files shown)


juanmicrosoft:

Issue #673, opened 2026-06-30
v0.7 — MCP agent-loop: doc-truth patch (0.6.6) + calor_scaffold spine (0.7.0)


juanmicrosoft:
Tracking issue for the v0.7 MCP agent-loop work. Plan: `docs/plans/v0.7-mcp-agent-loop.md` (PR #672), finalized after a 5-loop multi-persona review.

## Why
The shipped MCP `primer`/`tags`/`manifest` resources teach removed closer-form syntax (`§/F`, `§/M`) that hard-errors with **Calor0830** (no autofix). An agent following Calor's own docs writes non-compiling Calor. The plan first makes the docs *compile* (and guards that with a semantic test), then ships one real new capability.

## Track 1 — 0.6.6 (patch; additive bug fixes only)
- [ ] **PR 1 (D1)** — remove closer-form from primer/tags/manifest; rewrite CORRECT examples as complete compilable modules; tag WRONG examples; ULID→compact prose; fix flat↔nested option invocations; repoint `convert includeEffects` step; retask `generate_ids`.
- [ ] **PR 2 (D1b)** — `Calor0830 LegacyCloserForm` carries a `SuggestedFix` that strips the closer span, so `autoFix` auto-heals an agent's reflexive closer (Parser.cs:224–234).
- [ ] **PR 3 (D2a)** — contract guard test: semantic snippet compilation (CORRECT⇒0 diagnostics, WRONG⇒errors) + ULID denylist + existing-key path-nesting + prose.
- [ ] **PR 4** — bump version to 0.6.6 (create-release skill).

## Track 2 — 0.7.0 (minor; <1.0.0 ⇒ pre-release)
- [ ] **PR 5 (B)** — `calor_help action=effects_for` (enum + `calls` schema; `ParseCallTargetForChain`→internal; `PureExplicit` status).
- [ ] **PR 6 (C)** — `includeEffectSummary` on `calor_compile` (top-level, hoisted `EffectSummary?`, pre-fix capture, structured `missing`).
- [ ] **PR 7 (S)** — `calor_scaffold` (minimal): passes-the-compiler module + `ids-only` mode; call-less `§E{}` stubs; **typed-default `§R`** (BOOL:false/INT:0, not empty §R); BCL effect-equality acceptance bar; subsumes `generate_ids`.
[...]

PR #676, opened 2026-06-30
fix(Calor0830): auto-heal legacy closers + correct broken remediation message

juanmicrosoft:
## Summary

Track 1 / 0.6.6 item **D1b** — make `Calor0830` (legacy closer form) **auto-healable** and stop its message from sending users to a command that can't fix the file.

### The bug

Closer form (`§/M`, `§/F`, `§/I`, `§/L`, …) was removed in Phase 4d and now hard-errors `Calor0830`. The diagnostic told users to **"Run `calor format`"** to rewrite the file in indent form. But `calor format` — and `calor lint --fix` — **parse the file first and abort on `HasErrors`** (`FormatCommand.FormatFileAsync` / `LintCommand.LintFileAsync`). Since `Calor0830` *is* a parse error, those commands can't even read the file, let alone fix it. The remediation was a dead end.

### The fix

- **Auto-heal via `SuggestedFix`.** `Parser.ReportLegacyCloser` now reports through `ReportErrorWithFix`, attaching a fix that deletes the entire closer line — the keyword plus any optional `{id}` payload (e.g. `§/F{f1}`), located by peeking the `{`/`}` tokens. The deletion leaves a blank line, which indent-only Calor tolerates, so the healed source compiles.
  - This flows to the real agent-facing surfaces: the **LSP quick-fix** ("Remove the legacy closer …") and the **`calor_check` MCP tool with `apply: true`**, which strips the closers and returns clean source. (Both consume `DiagnosticBag.DiagnosticsWithFixes`; the parser produces the AST + fixes even though the parse "has errors".)
- **Corrected message.** `Calor0830` now explains the block ends at its body's dedent and to delete the closer line — no broken command reference.
- **Corrected stale doc comments** in `Diagnostic.cs` and `LegacyCloserFormLint.cs` that also claimed `calor format` strips closers automatically.

### Tests

- Parser-level (`ParserLegacyCloserRejectionTests`): the diagnostic carries a deleting `SuggestedFix`; the message no longer names `calor format`; applying the fixes to a closer-form module yields source that re-parses clean with no `Calor0830`.
- End-to-end (`DiagnoseToolFixTests`): `calor_check action=diagnose apply=true` on a closer-form module strips the closers and the healed source re-diagnoses to `errorCount: 0`.
- Full **`Calor.Compiler.Tests` (5532 passed)** and **`Calor.LanguageServer.Tests` (382 passed)** green; build clean under `TreatWarningsAsErrors`.

### Note / follow-up

There is still **no CLI command** that heals closer-form (`calor format` / `calor lint --fix` remain parse-first). The agent loop is covered by the MCP `calor_check apply` path and the LSP quick-fix. Wiring the existing source-level `LegacyCloserFormLint.Scan` (which already computes exact removal ranges, no parse needed) into a CLI remediation is a reasonable follow-up.

Relates to #673.

### DIFF
--- a/src/Calor.Compiler/Analysis/LegacyCloserFormLint.cs
+++ b/src/Calor.Compiler/Analysis/LegacyCloserFormLint.cs
@@ -10,9 +10,13 @@ namespace Calor.Compiler.Analysis;
 /// flagged.
 ///
 /// This is a SOURCE-level scanner (not AST) so it can be applied to
-/// any <c>.calr</c> source without parsing. The recommended machine
-/// fix is to run <c>calor format</c>, which re-emits the entire file
-/// in canonical indent form.
+/// any <c>.calr</c> source without parsing — useful precisely because
+/// closer form hard-errors (<c>Calor0830</c>) during parsing, which
+/// prevents the AST-based <c>calor format</c> / <c>calor lint --fix</c>
+/// paths from healing such files. Each <see cref="Finding"/> carries the
+/// exact source range to delete (<see cref="Finding.RemovedOffset"/> /
+/// <see cref="Finding.RemovedLength"/>); removing those ranges rewrites
+/// the source into canonical indent form.
 ///
 /// Wiring into the standard <c>calor lint</c> pipeline is intentionally
 /// gated: callers must request the lint explicitly so repositories
--- a/src/Calor.Compiler/Diagnostics/Diagnostic.cs
+++ b/src/Calor.Compiler/Diagnostics/Diagnostic.cs
@@ -294,17 +294,21 @@ public static class DiagnosticCode
     public const string CompactIdCollision = "Calor0822";
 
     /// <summary>
-    /// Info (opt-in lint, Phase 4b): a structural closing tag (e.g.
-    /// <c>§/F</c>, <c>§/CL</c>, <c>§/L</c>) is present in source that
-    /// has otherwise adopted indent form. Indent form alone now
-    /// terminates every structural block; legacy closers are kept only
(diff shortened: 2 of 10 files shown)


juanmicrosoft:
Follow-up commit 7b50d023 fixes a corruption bug discovered during confidence review.

**Bug:** Auto-healing a **bare** legacy closer (`§/F` with no `{id}` — the common modern form) produced scrambled output. The parser fired BOTH `Calor0830` (delete the closer line) AND `Calor0101` (id mismatch, whose fix *inserts* the opener id back into the closer). Applying both conflicting fixes scrambled the source. Id-bearing closers (`§/F{f1}`) accidentally avoided this because the echoed id matched the opener, so `Calor0101` never fired — which is why the original tests (all id-bearing) passed.

**Fix:** `ShouldReportMismatchedId` now returns `false` for any `StructuralLegacyClosers` token. Phase 4d removed closer form entirely, so `Calor0830` already owns these tokens and its fix deletes the whole line; an id mismatch on a doomed closer is moot and its insert-fix only conflicts.

Modernized 6 tests that asserted the obsolete `MismatchedId`/"does not match" behavior on structural closers (`§/M`, `§/F`, `§/L`, `§/CL`, `§/EACHKV`) to assert `Calor0830` + `DoesNotContain MismatchedId`. `§/TR` (try closer, not a structural closer) keeps its mismatch test unchanged.

Full suites: Calor.Compiler.Tests 5533 passed, Calor.LanguageServer.Tests 382 passed, 0 failed. Build clean.

PR #677, opened 2026-06-30
test(mcp): assert primer Common mistakes examples don't compile (D2a)

juanmicrosoft:
**PR 3 (Track 1, item D2a)** — the semantic guard that completes the primer-compiles work.

Stacked on #674 (depends on its compilable primer + the "Common mistakes" section). **Base = `fix/mcp-primer-compiles`**; review/merge #674 first.

### What
`PrimerMistakesRejectedTests` is the dual of #674's `PrimerCompilesTests`:
- PrimerCompilesTests proves every **CORRECT** module the primer teaches compiles.
- This proves every example in the primer's **"Common mistakes (these do NOT compile)"** section genuinely fails to compile — so the docs can't lie in the other direction (a "wrong" snippet that silently compiles would mis-teach an agent following Calor's own primer).

### How (fragments → complete modules → two-layer check)
The mistakes are fragments, so each is rewritten into the smallest complete module where it would naturally appear, then checked for non-compilation at **either** layer:
1. Calor reports an error (`HasErrors`), or
2. the generated C# fails Roslyn compilation.

The 4-field `§F{f1:Add:i32:pub}` header is the interesting one: Calor **accepts** it but emits `void Add() { return 0; }` → **CS0127** ("drops the return type"), caught only at the C# layer. (Follow-up candidate: a Calor-level "value returned from void function" check — the plan's "F-prerequisite invariant" — would catch this at layer 1; deferred to keep 0.6.6 test-only.)

### Drift guards (test ⇄ docs stay in sync)
- `Primer_ListsEachCuratedMistake` — every curated fragment must still appear in the primer block.
- `Primer_MistakeCount_MatchesCuratedSet` — adding a primer mistake without a curated case fails CI.
- `CorrectModule_CompilesAtBothLayers` — sanity so a green can't be an artifact of the Roslyn helper rejecting everything.

### Verification
Full `Calor.Compiler.Tests` green: **5546 passed, 0 failed**. New tests: 9 (6 mistake cases + 3 guards). Build clean (`TreatWarningsAsErrors`).

Linked to #673.

### DIFF
--- a/tests/Calor.Compiler.Tests/Mcp/PrimerMistakesRejectedTests.cs
+++ b/tests/Calor.Compiler.Tests/Mcp/PrimerMistakesRejectedTests.cs
@@ -1,0 +1,185 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Calor.Compiler;
+using Calor.Compiler.Effects;
+using Calor.Compiler.Mcp;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Xunit;
+
+namespace Calor.Compiler.Tests.Mcp;
+
+/// <summary>
+/// Semantic guard for the <c>calor://primer</c> MCP resource (v0.7 plan, item D2a) — the
+/// dual of <see cref="PrimerCompilesTests"/>. PrimerCompilesTests proves every CORRECT
+/// module the primer teaches actually compiles; this proves every example in the primer's
+/// "Common mistakes (these do NOT compile)" section genuinely FAILS to compile, so the docs
+/// cannot lie in the other direction (a "wrong" snippet that silently compiles would
+/// mis-teach the agent following Calor's own primer).
+///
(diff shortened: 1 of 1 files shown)


juanmicrosoft:
ℹ️ **CI / base note:** this PR targets `fix/mcp-primer-compiles` (it depends on #674's compilable primer + "Common mistakes" section). The repo's CI workflows (`test.yml` etc.) only trigger on PRs whose **base is `main`**, so the checks list here will stay empty until this branch is rebased onto `main` after #674 merges. Base was intentionally kept as #674's branch (not `main`) so this PR can't merge #674's changes ahead of #674's own review.

Local verification on this branch: `dotnet test Calor.Compiler.Tests` → **5546 passed, 0 failed**; build clean under `TreatWarningsAsErrors`. After #674 squash-merges, rebase onto `main` and CI will run the test-only diff.

juanmicrosoft:
Rebased onto main (dropping #674's now-merged commit) and retargeted base to \main\. Diff is now test-only (\PrimerMistakesRejectedTests.cs\, +185). Verified locally against merged main: PrimerMistakes + PrimerCompiles = 17/17 pass. CI now runs.

PR #721, opened 2026-07-22
feat(self-check): compile-check the agent syntax exemplar (#712)

### LINKED ISSUE #712: Snippet-level parse/compile check for exemplar idiom lines
PR #707 review residual: the E1a bug class (List<str> vs [str] type errors in exemplar lines) is not machine-guarded — keyword scan checks §-tags not types, and fragment fences are untagged so ParseExampleDocs skips them. §13 promises parse-checked exemplars; needs snippet-level parse or compile checking for non-§M fragment lines. Mutation test to beat: reintroducing the List<str> ReadAllLines bug must fail self-check.

juanmicrosoft:
Closes #712.

## The gap

The agent syntax exemplar (`Resources/agent-syntax-exemplar.md`, served to agents verbatim as `calor://primer`) states: *"Every complete `§M{...}` program below compiles (the compiler's own tests prove it)."* Nothing enforced that:

- The doc-drift example check (`Calor1328`) only fires on ` ```calor `-tagged fences; the exemplar used **bare** ` ``` ` fences, so every block was skipped.
- Even when scanned, that check only **parses**.
- No test compiled the exemplar at all.

And the failure mode it needed to catch is invisible to both parse-checking and Calor's own compile. I verified this directly: `§B{lines:List<str>} §C{File.ReadAllLines}` **parses cleanly, and `calor -i` emits it with exit 0** — producing `List<string> lines = File.ReadAllLines(path);`. Only the C# compiler rejects it (`CS0029`, plus `CS1061` on `.Length`). That is the E1a trap that shipped into agent-authored code.

## The fix

New `ExemplarCompileChecker`, wired into `calor self-check docs`:

1. **Complete programs → full C# compile (`Calor1330`).** Each complete `§M` program is emitted to C# and the **generated C# is compiled through Roslyn's full semantic model** — the only layer that catches type errors. The SDK's implicit global usings are supplied so `File`/`List`/etc. resolve (the emitter relies on `ImplicitUsings`).
2. **Fragment reference lines → targeted lint (`Calor1331`).** The copyable "Calls"/"Lists" fragments intermix prose and free identifiers (`path`, `Obj`, `LocalFunc`), so they cannot be compiled standalone. The one recurring trap they can carry is linted directly: an array-returning BCL call (`File.ReadAllLines`, `Directory.GetFiles`, …) must bind to the array form `[T]`, never a generic collection.
3. **Tagged the 9 complete-program fences ` ```calor `** so they are unambiguously identified. Reference blocks like "Core tags" — whose first line is the shape line `§M{id:Name}   Module` — stay bare ` ``` ` and are correctly excluded.

## Mutation test (the acceptance criterion)

Reintroducing the `List<str>` `ReadAllLines` bug **fails self-check** wherever it lands:

| Mutation site | Caught by |
|---|---|
| Complete program (`m3:Files`) | `Calor1330` — Roslyn `CS0029: cannot convert string[] to List<string>` (+`CS1061`) **and** `Calor1331` |
| Fragment line ("Calls" block) | `Calor1331` — the array-trap lint (cannot be compiled standalone) |

Both drive `self-check docs` to a nonzero exit. The clean exemplar stays green.

## Honest scope
Arbitrary fragment lines are **not** compiled — they reference undeclared identifiers and abstract placeholders (`§C{Obj.Method}`, `§C{LocalFunc}`), so wrapping-and-compiling them generically produces false positives. The deep compile covers the complete programs (which the file's own contract says exercise every idiom); the lint covers the single documented trap those fragments repeat. This is called out in code comments and `docs/cli/self-check.md`.

## Changes
- `SelfCheck/ExemplarCompileChecker.cs` (new) — extract programs, Calor→C→Roslyn semantic compile, array-trap lint.
- `SelfCheck/DocDriftChecker.cs` — new `ExemplarDoc` input, wired into `Check`/`LoadFromRepository`.
- `Diagnostics/Diagnostic.cs` — `Calor1330`/`Calor1331`; band comment extended to 1320–1331.
- `Resources/agent-syntax-exemplar.md` — 9 complete-program fences tagged `calor`.
- `docs/cli/structured-output.md`, `docs/cli/self-check.md`, command help — document the two new codes and the deep check.
- `tests/.../ExemplarCompilesTests.cs` (new) — every program compiles; both mutation paths caught; array form accepted; program count == 9. Runs in every CI environment.

## Verification
- `Calor.Compiler.Tests` — 5827 passed (incl. 5 new), 7 skipped.


juanmicrosoft:
Thanks — this was the sharpest review of the set. Pushed fixes for the three merge-blockers plus the cheap correctness items:

**1. Honor `drift:ignore` (demonstrated defect).** Both new checks now consult the same preceding-line marker the rest of self-check uses. The array-trap lint skips a flagged line whose previous line carries `<!-- drift:ignore -->`, and the complete-program compile skips a fenced block whose opening fence is so marked. New test `ArrayTrapLint_HonorsDriftIgnoreMarker` locks it in; verified end-to-end that a "do NOT write this" negative example is now exempt.

**2. Parse-level double-report (demonstrated).** The exemplar is now excluded from `ParseExampleDocs`, so `Calor1328` no longer fires alongside `Calor1330`. The deep compile's stage 1 already reports any parse failure, so rot surfaces exactly once (verified: the missing-`id` `§IF` case now yields a single `Calor1330`, and each program is parsed once).

**4. Characterization test no longer pins the bug.** `ReadAllLinesBoundToList_...` now asserts the trap is caught at the Calor **or** the C# layer (`calorCaught || roslynCaught`) with a comment pointing at #722 — a binder fix flips `calorCaught` true without reading as a regression.

**5. Wrong-layer / oversell.** Filed **#722** for the real fix (binder rejecting an array-typed initializer bound to `List<T>`, so `calor -i` fails instead of emitting broken C#). Softened the CHANGELOG: it now says plainly this guards the exemplar *document* and that the language-level fix is tracked separately. You're right that an agent writing the line independently still gets no protection today — that's #722's job.

**6. Minors.** Array-API list carries an "extend here" comment noting it's a deliberately closed list (the complete programs get the real compile for any API); `GlobalUsingsPreamble` now pins which SDK's `ImplicitUsings` set it mirrors and notes the failure mode if they drift; the reference cache is now `Lazy<T>` (thread-safe under xUnit parallelism). Stage-2 first-line anchoring left as-is per your note.

**3. Three parallel implementations — deferred as a fast-follow.** You're right they overlap. I left `PrimerCompilesTests`/`PrimerMistakesRejectedTests` in place for this PR because they aren't *strictly* subsumed — `PrimerCompilesTests` compiles under the MCP tool's `UnknownCallPolicy.Strict` options (mine uses `Program.Compile` defaults), so retiring it would drop that fidelity, and both extractors already carry their own count-guards. Consolidating onto the new TPA-referenced `RoslynErrors` and a single extractor is worth doing but is pure test-refactor with its own risk surface; happy to do it as a follow-up rather than widen this PR.

Build 0 warnings; Calor.Compiler.Tests 5823 passed (6 exemplar tests) / 7 skipped; `self-check docs` clean.

PR #723, opened 2026-07-22
feat(binder): reject array-typed initializer bound to a concrete collection (#722)

### LINKED ISSUE #722: Binder accepts array-typed initializer bound to List<T> (the real E1a fix, below #712)
## The real E1a fix: the binder should reject `[T]`-returning call bound to a generic collection

Follow-up to #712 (PR #721). #712 guards the *exemplar document* against the E1a array-vs-list trap, but the root cause is a level below: **`calor -i` itself accepts the bug and emits broken C#.**

Reproduce:

```
§M{m:Files}
  §F{f:CountLines:pub} (str:path) -> i32
    §E{fs:r}
    §B{lines:List<str>} §C{File.ReadAllLines} §A path §/C
    §R (len lines)
```

`calor -i bad.calr -o bad.g.cs` exits **0** and emits:

```csharp
List<string> lines = File.ReadAllLines(path);  // CS0029
return lines.Length;                            // CS1061 (List has no .Length)
```

The type error only surfaces when the *generated* C# hits Roslyn. Any agent who writes this line independently — not just one copying the exemplar — gets a clean Calor compile and a failing `dotnet build` downstream. That is the actual E1a failure mode; #712's doc guard only covers the one file that might teach it.

## Ask

Teach the binder / type checker to reject an initializer whose type is an array (`[T]`, e.g. the result of `File.ReadAllLines`/`Directory.GetFiles`) when the binding declares an incompatible generic collection type (`List<T>`, `IList<T>`, …), with a dedicated diagnostic (Calor02xx semantic band). Mirror the C# assignability rule: array → `List<T>` is not an implicit conversion.

## Acceptance
- The program above fails `calor -i` with a semantic diagnostic pointing at the binding, not a downstream CS0029.
- `ExemplarCompilesTests.ReadAllLinesBoundToList_FailsTheRoslynSemanticCompile` (which currently characterizes Calor *missing* this) is updated to assert Calor now catches it.
- Broader than ReadAllLines: covers array/collection assignability generally, not a closed API list.

## Context
- PR #721 investigation proved the exit-0 behavior.
- #712 / PR #721 is the docs-level mitigation; this is the language-level fix.

juanmicrosoft:
Closes #722.

## The bug

`calor -i` accepted the E1a trap and emitted broken C#:

```
§B{lines:List<str>} §C{File.ReadAllLines} §A path §/C
```
→ exit **0**, emitting `List<string> lines = File.ReadAllLines(path);` — which fails the downstream `dotnet build` with **CS0029** (`string[]` is not implicitly convertible to `List<string>`). #712 guarded the exemplar *document*; this fixes the language so **any** agent who writes the line — not just one copying the exemplar — gets a clean, local error.

## The fix

The diagnostic lives in **`BindValidationPass`**, the always-on AST pass that already owns `§B` validation. (Discovery worth noting: the `Binder`/bound-tree is **not** in the default `calor -i` path — compilation is AST→emit — so the check had to go in an AST pass, not the binder proper.)

`Calor0254 BindArrayToConcreteCollection` fires when a binding declared as a **concrete generic collection** (`List`, `HashSet`, `SortedSet`, `Queue`, `Stack`, `LinkedList`, `Collection`, `ObservableCollection`, `SortedList`) is initialized by an **array**. It mirrors C#'s actual rule:

| Declared type | Array initializer | Result |
|---|---|---|
| `List<str>` / `HashSet<str>` / … | yes | **Calor0254** (CS0029) |
| `[str]` (array form) | yes | ok |
| `IEnumerable<str>` / `IList<str>` (interface) | yes | ok — arrays satisfy the interfaces |
| `List<str>` | non-array | ok |

The initializer is recognized as an array when it calls a known array-returning BCL method (`File.ReadAllLines`/`ReadAllBytes`, `Directory.GetFiles`/`GetDirectories`/`GetFileSystemEntries`) **or a user function declared `-> [T]`** — return types are collected per module, so the check generalizes beyond a closed API list on the collection side and to user code on the source side.

```
$ calor -i bad.calr -o bad.g.cs
bad.calr(4,5): error Calor0254: Binding 'lines' declares 'List<STRING>', but its
initializer is an array. An array is not implicitly convertible to the concrete
collection 'List<…>' (the emitted C# would fail with CS0029). Use the array form
'[str]', or wrap the value explicitly (e.g. a new List constructed from it).
```

## Acceptance (#722)
- ✅ The issue's repro fails `calor -i` with a semantic diagnostic at the binding, not a downstream CS0029.
- ✅ `ExemplarCompilesTests.ReadAllLinesBoundToList_...` updated to assert Calor now catches it (`Calor0254`), with the Roslyn compile kept as a backstop.
- ✅ General on the collection side (any concrete collection) and the array side (BCL registry + user functions returning `[T]`), not a single hardcoded API.

## Tests
- **`ArrayToCollectionBindTests`** — BCL and user-function array sources rejected; array form and collection **interfaces** accepted; not-an-array initializer accepted; `HashSet` covered.
- **`BindCorpusCleanTests`** now pins `Calor0254` clean across `samples/` + benchmarks (grep confirmed no corpus file binds a collection to an array method).
- `docs/syntax-reference/binding.md` documents `Calor0254` as an always-on hard error, distinct from the strict-inference trio (which fire on the *absence* of a type; this fires *because of* the explicit type).

## Verification
- Compiler.Tests 5829 / Semantics 37 / Evaluation 199 / Conversion 286 / Enforcement 271 — all green.


juanmicrosoft:
Thanks — all six addressed.

**1. Trap only caught in binding position (demonstrated).** Scoped honestly rather than overclaiming: the CHANGELOG now says "covers **binding** position" and filed **#724** for `§ASSIGN` / return / argument (with the concrete infra each needs — target-var type map, threaded return type, call-site type flow). You're right that the exemplar's own trap text is about signatures; #724 calls that out. Kept this PR tight per your steer rather than growing it.

**2. "Always-on" had no unit test.** Added `IsAlwaysOn_EvenWithStrictInferenceDisabled` — same repro with `strictInference: false`. A future refactor that folds the check into the strict-gated block now fails a test instead of silently going opt-in.

**3. Tautological backstop assertion.** Fixed. The exemplar test now asserts only the Calor layer for binding position. The backstop gets its own test, `ReturnPositionArrayToList_IsCaughtByTheRoslynBackstop`, using a genuine binder-blind input (return position, which #722 doesn't cover): Calor compiles clean, the emitted `List<str>`-returning method that returns `string[]` fails Roslyn with CS0029. Real coverage, and it documents the #724 gap.

**4. Two hand-synced tables in a drift-checking compiler.** Extracted `Analysis/ArrayReturningBcl` as the single source of truth. `BindValidationPass` reads it directly; `ExemplarCompileChecker` builds its lint regex from the same keys. There is no longer a pair to sync — and `DocsAndCompilerGuards_ShareOneBclApiList` pins that both guards recognize every API in the shared table.

**5. BCL table shadowed user names.** Swapped the order — `_userReturnTypes` is consulted before `ArrayReturningBcl`, so a user `File.ReadAllLines -> List<str>` is judged by its real return type (no false `Calor0254`). `UserFunctionShadowingBclName_IsJudgedByItsRealReturnType` covers it. (The bare-name/`Class.Method` "last write wins" remains a documented rare-collision heuristic feeding a conservative check.)

**6. Minors.** `ReadAllBytes → u8` verified correct (types.md maps `u8`→`byte`); cross-module false-negative noted in a comment. The normalized-type echo in the message (`List<STRING>`) is left as-is — accurate but cosmetically inconsistent with the `[str]` suggestion; not worth a normalized→source mapping in this PR.

Build 0 warnings; Compiler.Tests 5833 / Semantics 37 / Evaluation 199 / Conversion 286 / Enforcement 271 green; `self-check docs` clean.

Issue #725, opened 2026-07-22
Extend Calor0254 array→collection check to argument position (call-site type flow)


juanmicrosoft:
## Extend Calor0254 to argument position (call-site type flow)

Split from #724, which explicitly sanctioned this: "Argument position may be split into its own issue if the call-site type-flow work is large." #724 (PR #TBD) covers binding, return, and assignment positions. Argument position remains:

```
§F{f:Take:pub} (List<str>:items) -> i32
  §R (len items)
§F{g:Use:pub} (str:path) -> i32
  §E{fs:r}
  §R §C{Take} §A §C{File.ReadAllLines} §A path §/C §/C   // array fed to a List<str> parameter
```

The argument `File.ReadAllLines` (an array) is passed where `Take` declares a `List<str>` parameter → CS0029 in the emitted C#, but `calor -i` accepts it.

## Why it's separate
Binding/return/assignment all check a single declared type against a single value expression the pass already visits. Argument position needs **call-site type flow**: a map of function/method name → parameter types, matched positionally against each call's argument expressions, at every call site (statement-level, initializer, return value, and calls nested inside other expressions). That means a general expression walker plus positional param matching — materially larger and with overload-resolution edge cases (Calor allows arity-based sibling overloads).

## Ask
- Build `name → parameter types` per module (mirror of `_userReturnTypes`).
- Walk expressions; for each `CallExpressionNode` to a known user function, check each argument via the existing `InitializerArrayElement` against the callee's parameter type using `TryGetConcreteCollectionName`.
- Emit `Calor0254` at the argument's span.
- BCL callees are skipped (no parameter-type registry) — a conservative false negative.

## Acceptance
- The reproduction above fails `calor -i` with `Calor0254` at the argument.
- `BindCorpusCleanTests` stays clean.

## Context
- Machinery to reuse: `Analysis/ArrayReturningBcl`, `TryGetConcreteCollectionName`, `InitializerArrayElement`, `_userReturnTypes` (add a parallel `_userParamTypes`).

PR #726, opened 2026-07-22
feat(binder): extend Calor0254 array→collection to return and reassign (#724)

### LINKED ISSUE #724: Extend Calor0254 array→collection check to §ASSIGN / return / argument positions
## Extend Calor0254 (array→concrete-collection) beyond binding position

Follow-up to #722 (PR #723). #722 catches the array-vs-collection trap only in **binding** position (`§B{x:List<str>} §C{File.ReadAllLines}`). The identical mistake still compiles with exit 0 and emits C# that fails `dotnet build` with CS0029 in three adjacent positions:

**Reassignment (§ASSIGN):**
```
§B{~items:List<str>} lines
§ASSIGN items §C{File.ReadAllLines} §A path §/C
```

**Return position (§R)** — this is the variant the exemplar's own trap text is about ("Do NOT copy [str] into signatures that require List<str>"):
```
§F{f:Get:pub} (str:path) -> List<str>
  §R §C{File.ReadAllLines} §A path §/C
```

**Argument position** — a parameter declared `List<str>` fed an array argument at a call site.

Each emits a `List<string>` context assigned/returned a `string[]` → CS0029.

## Ask
Reuse the #722 machinery (`ArrayReturningBcl`, `TryGetConcreteCollectionName`, the per-module user-return-type map) to flag the same trap in these three positions, all emitting `Calor0254`:
- **§ASSIGN**: needs the target variable's declared type (a per-function `name → declared type` map built from `§B` statements).
- **§R**: needs the enclosing function/method's declared return type threaded through the statement walk.
- **argument**: needs the callee's parameter types matched against call-site argument expressions (the largest piece — call-site type flow).

## Acceptance
- The three reproductions above fail `calor -i` with `Calor0254` at the offending statement.
- `BindCorpusCleanTests` stays clean.
- Argument position may be split into its own issue if the call-site type-flow work is large.

## Context
- PR #723 review demonstrated all three exit-0 on the #722 head.
- `ExemplarCompilesTests.ReturnPositionArrayToList_IsCaughtByTheRoslynBackstop` currently documents the return-position gap (caught only by the Roslyn backstop); flip it to assert the Calor layer once §R is covered.

juanmicrosoft:
Closes #724. (Argument position split to #725.)

## What

#722 caught the array-vs-concrete-collection trap (`Calor0254`) only in **binding** position. The identical mistake still compiled with exit 0 and emitted C# that fails `dotnet build` with CS0029 in adjacent positions. This extends the check to two more:

**Return (`§R`)** — the signature variant the exemplar's own trap text warns about:
```
§F{f:Get:pub} (str:path) -> List<str>
  §R §C{File.ReadAllLines} §A path §/C     // Calor0254
```

**Reassignment (`§ASSIGN`):**
```
§B{~items:List<str>}
§ASSIGN items §C{File.ReadAllLines} §A path §/C   // Calor0254
```

## How

`BindValidationPass` now threads **per-body context**, reset per function-like member:
- `_currentReturnType` — the body's declared return type, for `§R` (functions, methods, operators, property/indexer getters; `null` for ctors/setters).
- `_localVarTypes` — a `name → declared type` map populated from `§B` statements as the body is walked, for `§ASSIGN` targets.

All three positions (binding, return, assign) share one `CheckArrayToCollection(value, declaredType, span, leadIn)` helper — same array-source detection (`InitializerArrayElement` → the shared `ArrayReturningBcl` table + user functions returning `[T]`), same collection-type detection (`TryGetConcreteCollectionName`), same rule (collection **interfaces** like `IEnumerable<T>`/`IList<T>` still accepted). The `Check` loop was also de-nested into a single `CheckBody(body, returnType)` call per member.

## Scope
**Argument position** (an array passed to a `List<T>` parameter) needs call-site type flow — a general expression walk plus positional parameter matching against every call site, with arity-overload edge cases. As #724 itself sanctioned splitting it, that's tracked in **#725**.

## Tests
- `ArrayToCollectionBindTests`: return + reassign rejected; array-form return type and array-form variable accepted (no over-fire).
- `ExemplarCompilesTests`: the return-position characterization test flips from "caught only by the Roslyn backstop" to asserting the **Calor** layer catches it (`Calor0254`).
- `BindCorpusCleanTests`: still clean — the two new positions add zero firings across `samples/` + benchmarks (the higher-traffic surface; verified no false positives).
- `docs/syntax-reference/binding.md` documents all three covered positions and points at #725 for argument.

## Verification
- Compiler.Tests 5837 / Semantics 37 / Conversion 286 / Enforcement 271 / Evaluation 199 — all green.


juanmicrosoft:
Thanks — the §EACH hole was a real one and I'm glad you probed it. All addressed.

**1. Blind inside `§EACH` bodies (blocking, demonstrated).** `CheckStatement` now recurses into every block-bearing statement kind — I audited the AST and added `§EACH`/`§EACHKV` (`ForeachStatementNode`) plus the ones that were also missing: `§USING`, `§SYNC`, unsafe, and fixed blocks. This restores the whole family (0251–0254) inside those bodies, not just the array check. Regression test `TrapInsideForeachBody_IsRejected` (trap in a `§EACH` over `Directory.GetDirectories` → `Calor0254`). Verified the exact idiom now errors.

**2. `§ASSIGN` to a class field / parameter uncovered (demonstrated).** Took the cheap-and-correct path: the per-body base scope is now seeded with the enclosing class's fields and the member's parameters (both statically declared — no type flow). A `List<str>` field reassigned an array is caught (`AssignToCollectionTypedField_IsRejected`), and the array-form field is accepted. Parameters shadow fields of the same name, matching resolution.

**3. Scope-blind flat map (demonstrated).** Replaced it with a proper lexical scope stack — each block pushes a child scope and pops on exit, lookup walks innermost-out. An inner-block binding can no longer overwrite a same-named outer variable's type. `InnerScopeShadow_DoesNotMistypeOuterAssign` pins it (outer `[str]` var, inner `List<str>` shadow, correct outer `§ASSIGN` → no false `Calor0254`). And I filed **#727** for the adjacent codegen hole your probe exposed: `calor -i` accepts local shadowing and emits CS0136 C# — a distinct bug in the same family, now tracked rather than resting on an unwritten invariant.

**4. Polish.**
- The message no longer leaks the normalized spelling: it's now `Binding 'lines' is a concrete collection ('List<…>'), but the value is an array … Use the array form '[str]' …` — one convention, agent-facing-clean, consistent across all three positions.
- `§LAM` non-traversal documented in `binding.md`'s scope note (a block lambda `-> List<T>` returning an array is unchecked).
- `_scopes`/`_currentReturnType` single-threaded/non-reentrant nature noted in a comment.

Build 0 warnings; Compiler.Tests 5841 / Semantics 37 / Conversion 286 / Enforcement 271 / Evaluation 199 green; `self-check docs` clean (it caught two bogus keyword spellings in my first scope-note draft — fixed).

Issue #727, opened 2026-07-22
calor -i accepts local shadowing that emits CS0136 C# (block-local shadows enclosing local)


juanmicrosoft:
## `calor -i` accepts local shadowing that emits CS0136 C#

Surfaced during the PR #726 review (finding 3). Same family as #722/#724/#725: `calor -i` exits 0 and emits C# that fails a downstream `dotnet build`.

A `§B` inside a nested block may reuse the name of a binding in an enclosing scope. Calor accepts it and emits two `var`s with the same name in nested scopes:

```
§M{m:Shadow}
  §F{f:Do:pub} (bool:flag) -> i32
    §B{~x:i32} 0
    §IF{if1} (== flag true)
      §B{x:str} "hi"
      §P x
    §R x
```

`calor -i` → exit **0**, emitting:
```csharp
int x = 0;
if (flag == true) {
    string x = "hi";   // CS0136: a local named 'x' cannot be declared in this scope
    Console.WriteLine(x);
}
return x;
```

C# forbids a nested local from shadowing an enclosing local (CS0136), so this fails `dotnet build`.

## Ask
Detect an inner `§B` whose name collides with an in-scope enclosing local (or parameter) and either:
- reject it with a dedicated diagnostic (Calor02xx) at the inner binding, or
- emit a uniquely-renamed local so the generated C# compiles (the more permissive option — but then references must be renamed consistently).

Rejecting is simpler and matches C#'s own rule.

## Acceptance
- The program above fails `calor -i` with a semantic diagnostic (or compiles to valid, CS0136-free C#).
- Corpus stays clean.

## Context
- Found while adding lexical scoping to `BindValidationPass` for #724 (PR #726). That pass now scopes correctly, so it does not itself mis-report on shadowed names — but the underlying codegen hole remains.

PR #728, opened 2026-07-22
feat(binder): extend Calor0254 array→collection to argument position (#725)

### LINKED ISSUE #725: Extend Calor0254 array→collection check to argument position (call-site type flow)

juanmicrosoft:
Closes #725. Completes the array-vs-concrete-collection trap class across all four positions.

## What

`Calor0254` now also fires when an array is passed where a user function/method declares a concrete generic collection parameter — the call-site type-flow case that #724 split out:

```
§F{f:Take:pub} (List<str>:items) -> i32
  §R (len items)
§F{g:Use:pub} (str:path) -> i32
  §E{fs:r}
  §R §C{Take} §A §C{File.ReadAllLines} §A path §/C §/C   // Calor0254 at the argument
```

```
$ calor -i arg.calr -o arg.g.cs
arg.calr(6,20): error Calor0254: Parameter 1 of 'Take' is a concrete collection ('List<…>'),
but the value is an array. … Use the array form '[str]', or construct a new List from the array.
```

## How

`BindValidationPass` gains:
- **`_userParamTypes`** — `name → parameter types`, keyed by `"name/arity"` (and `"Type.Method/arity"`), so Calor's arity-based sibling overloads resolve to the correct signature.
- **`ScanExpressionForCalls`** — a recursive walker that finds every call reachable from a checked expression (a call's own arguments, plus binary/unary/conditional/type-op operands) and checks each argument against the callee's declared parameter type via the existing `InitializerArrayElement` + `TryGetConcreteCollectionName`.
- Call scanning wired into **every expression position**: `§B` initializers, `§R`/`§ASSIGN` values, `§C` call statements, `§P`/expression statements, and `if`/`while`/`do-while` conditions + `§EACH` collections.

## Honest limits (documented in binding.md)
- Only **same-module user callees** are resolved — BCL and cross-module callees have no parameter-type registry (conservative false negatives).
- Block-lambda (`§LAM`) bodies are still not traversed.
- Unhandled expression node types simply stop the descent — a safe false negative, never a false positive.

## Tests
- `ArrayToCollectionBindTests`: array arg → `List<str>` param rejected (nested in a return **and** as a call statement); array-form param accepted; **overload arity disambiguation** (the 1-arg `List<str>` overload rejects, the 2-arg `[str]` overload accepts, resolved purely by call arity).
- `BindCorpusCleanTests`: still clean — the whole-corpus argument scan adds zero firings (verified no false positives on the higher-traffic surface).

## Verification
- `dotnet build` — 0 warnings.
- Compiler.Tests 5844 / Semantics 37 / Conversion 286 / Enforcement 271 / Evaluation 199 green. (One `WatchDebouncerTests` timing flake in a full-suite run passed on isolated re-run — unrelated to this change.)

With #722/#724/#725, the array-vs-list trap is caught at the language level in binding, return, reassignment, and argument positions. Remaining adjacent hole: local shadowing → CS0136 (#727).


juanmicrosoft:
Thanks — finding 1 was a real one, and the fix turned out to make both directions correct rather than trading one FN for another.

**1. Hard-error false positive (blocking, demonstrated).** Root-caused and fixed. Methods were registered under both `Class.Member/arity` **and** the bare `Member/arity`, after free functions — so a method clobbered a same-name/arity free function's params and raised `Calor0254` on valid code. Now methods are stored **only** under `Class.Member`, free functions under the bare name, and unqualified calls resolve **context-sensitively**: a call inside class `C` prefers `C`'s member (implicit `this`), otherwise a module-level free function. Your exact probe now compiles clean (`FreeFunctionSharingMethodName_IsNotClobbered`). I applied the same resolution to **return-type** lookup too — it had the identical latent clobber that could have produced a false positive the other way (free func `-> List<str>` vs method `-> [str]`).

**2. Clobbering asymmetry.** Gone — this is the part I'm happiest about. Because each callee is keyed distinctly and picked by context, there's no single flat-map race: the free-function call **and** the intra-class implicit-`this` call are **both** correct, neither a conservative FN. `IntraClassImplicitThisCall_ResolvesToTheMethod` pins the mirror direction. Documented the resolution rule in `binding.md`.

**3. Inconsistent statement expression positions.** Swept them: `match` subject, `using` resource, `sync` lock expression, and `fixed` initializer are now scanned (the `if` condition was already covered — added a test for a call in an `§IF` condition). Not "the ones that came to mind" this time — I walked every body-bearing / expression-bearing statement node.

**4. Smaller.**
- Constructors, operators, and indexers are **not** registered — now stated explicitly in `binding.md` as conservative FNs (rather than the PR body reading broader than the code).
- `p.TypeName` guarded with `?? ""` and a comment noting the parser's non-null invariant, so a future nullable-annotation change can't feed null into `.Trim()`.
- Filed **#729** for the residual `WatchDebouncerTests` wall-clock flake — the injectable-delay conversion `#716` started but didn't finish. (It's the one that failed a full-suite run here and passed on isolated re-run; unrelated to this change.)

Build 0 warnings; Compiler.Tests 5847 (my 24 array-trap tests + corpus-clean deterministic) / Semantics 37 / Conversion 286 / Enforcement 271 / Evaluation 199 green; `self-check docs` clean.

Issue #729, opened 2026-07-22
Finish converting WatchDebouncerTests off wall-clock timing (residual flakes)


juanmicrosoft:
## Finish converting WatchDebouncerTests off wall-clock timing (residual flakes)

Flagged in the #716 review and re-sighted across this PR series: `WatchDebouncerTests` intermittently fails a full-suite run and passes on isolated re-run — e.g. `EventWithinQuietPeriod_ExtendsBatch_AndQuietPeriodDurationIsHonored`, and the burst tests (`Burst_…`, `SeparateBursts_…`). These are the residual wall-clock-dependent tests; under load (a full `dotnet test`) the real-time timers race.

`#716` introduced the `TimeProvider`/`FakeTimeProvider` seam for the debouncer, and some tests were converted. The remaining ones still sleep on real time. Convert them to drive the injected `FakeTimeProvider` so the batching/quiet-period assertions are deterministic regardless of host load.

## Acceptance
- No `WatchDebouncerTests` case depends on wall-clock timing.
- The suite passes deterministically under parallel load (no "passed on isolated re-run").

## Context
- Machinery already exists (`WatchDebouncer` takes a `TimeProvider`); this is finishing an in-progress conversion, not new design.
- Sighted during PR #723/#726/#728 full-suite runs.

PR #730, opened 2026-07-22
feat(binder): reject local shadowing that emits CS0136 C# (#727)

### LINKED ISSUE #727: calor -i accepts local shadowing that emits CS0136 C# (block-local shadows enclosing local)

juanmicrosoft:
Closes #727.

## The bug

Same family as #722/#724/#725 — `calor -i` exited 0 on a `§B` whose emitted C# fails `dotnet build`. A binding declaring a **new** local that reuses the name of a local or parameter already in an **enclosing** scope emits a nested `T x = …` inside an outer `T x = …` → **CS0136**:

```
§B{~x:i32} 0
§IF{i1} (== flag true)
  §B{x:str} "hi"     // was accepted → emitted `string x = "hi";` inside `int x = 0;` → CS0136
```

## The fix

`BindValidationPass` (which already carries the lexical scope stack from #724) now reports `Calor0255`. The check **mirrors the emitter's exact decision** — this was the crux:

- **`CSharpEmitter` treats a mutable `§B{~x}` reusing a name already declared in the function as a *reassignment* (`x = …`), not a re-declaration.** So a mutable rebind is not shadowing — the accumulator idiom is valid and must not fire:
  ```
  §B{~result} 1
  §WH{w1} (<= i n)
    §B{~result} (* result i)   // emitter → `result = result * i;` — valid, no Calor0255
  ```
  A flat `_declaredInFunction` set mirrors `CSharpEmitter._declaredVariablesInCurrentScope` to make the same classification.
- **Fields are kept out of the scope stack** — a local may legally shadow a field (the local wins in C#), so only enclosing locals/parameters trip the check.
- **Parameters live in a base scope enclosing the body**, so a body local reusing a parameter name is caught.
- **Sibling (non-nested) blocks may reuse a name** — scopes pop on block exit.

## Corpus & a latent test bug

Adding `Calor0255` to `BindCorpusCleanTests` initially surfaced 11 firings — all mutable-rebind accumulators (Factorial/Fibonacci/GCD). Those are reassignments, and once the emitter-mirroring rule was in place the corpus is **clean**.

One MCP test, `TypeCheckToolTests.ExecuteAsync_ValidComplexSource_NoErrors`, asserted the validity of a source that used **immutable** `§B{result}` shadowing across branches — which actually emits CS0136. It was mis-asserting (Calor said OK; `dotnet build` would have failed). Corrected to the mutable accumulator form it clearly intended, with a comment.

## Tests
`LocalShadowingTests`: immutable inner shadow **rejected**; mutable rebind **accepted**; sibling reuse **accepted**; field shadow **accepted**; parameter reuse **rejected**. `BindCorpusCleanTests` now pins `Calor0255` clean.

## Verification
- `dotnet build` 0 warnings; `calor self-check docs` clean.
- Compiler.Tests (50 relevant tests deterministic) / Semantics 37 / Conversion 286 / Enforcement 271 / Evaluation 199 / Verification 253 green. (Intermittent full-suite failure is the `WatchDebouncerTests` timing flake tracked in #729 — passes on isolated re-run.)


juanmicrosoft:
Thanks — probe C (loop variables) was a real miss of the PR's own titled class, and finding 4 is the point that matters most. Addressed.

**1. CS0136 through loop variables (blocking, demonstrated).** Fixed. `§L`/`§EACH`/`§EACHKV` iteration variables are now pushed into a scope enclosing the body (`CheckLoopBlock`), so an inner `§B` reusing the iteration variable's name is caught. `§L{l1:i} … §B{i}` → `Calor0255`. Test added (`InnerBindingReusingALoopVariableName_IsRejected`).

**4. The mirror invariant is a comment, not a contract (the durable fix).** Added **`ShadowingDifferentialTests`**. For every idiom it asserts the real spec — **if `calor -i` accepts a program, its emitted C# compiles under Roslyn** — using `ExemplarCompileChecker.RoslynErrors` as ground truth. That's agreement with Roslyn, not with the emitter. It covers the accepted-clean idioms (accumulator, sibling-immutable, field-shadow, loop-ok), the correctly-rejected ones, and pins the three known gaps below with their exact CS codes. It would have caught all four of your probes, and any *new* accepted-but-broken idiom now fails a test instead of shipping.

**2 & 3. The other three holes — filed, not certified.**
- **#731** — same-scope duplicate `§B` (CS0128). I *tried* the broadening and reverted it: it flagged legitimate converter output — the C#→Calor converter emits array/list/dict reassignments as same-name creation blocks (`ConverterQualityTests`). So CS0128 genuinely needs the converter to emit reassignments as `§ASSIGN`/rebind, not a blanket reject. That's #731's scope.
- **#732** — sibling mutable rebind → CS0103. You nailed the structural point: the pass mirrors the emitter's *flat* `_declaredVariablesInCurrentScope`, so it certifies the emitter's bug. The fix is scope-aware tracking in **both** the emitter and the pass; filed with that plan.
- **#733** — type-changing mutable rebind → CS0029.

All three are pinned in `ShadowingDifferentialTests` as `KnownGap_*` facts asserting the current CS0103/CS0029/CS0128 output, each referencing its issue — so the map is executable and each fix will flip its assertion.

**5.** Documented the loop-variable coverage and the three gaps in `binding.md` and the `Calor0255` doc; commented the empty-param-type flow.

You named the pattern exactly: each PR's lane is solid and the edges are the next construct over. The differential test is the loop-closer — from here, an accepted-but-broken idiom is a red test, not a reviewer find.

Build 0 warnings; `self-check docs` clean; 87 relevant tests deterministic; full Compiler.Tests 5865 green.

Issue #714, opened 2026-07-21
Flaky: WatchDebouncerTests.EventsWithinQuietPeriod_ExtendTheSameBatch (real-time race)


juanmicrosoft:
## Problem
`WatchDebouncerTests.EventsWithinQuietPeriod_ExtendTheSameBatch` has flaked on **three consecutive PRs** (#703 release, #707, and observed on the #713 run before passing on rerun). Each flake costs a CI rerun. Per the gates-doc philosophy a flaky check is a protocol defect, not a nuisance.

## Root cause
The debouncer measures a real wall-clock quiet period: `WatchDebouncer.ReadBatchAsync` uses `Task.Delay(quietPeriod)` (WatchDebouncer.cs:42). The test races a real `await Task.Delay(20)` (WatchCommandTests.cs:43) against that timer to assert a second event joins the same batch. Under CI scheduling load the 20ms nudge can land after the quiet timer elapses, so event `b` lands in the *next* batch — `Assert.Contains("/p/b.calr", batch)` then fails with 'Item not found in set: ["/p/a.calr"]'. It is timing, not logic — it passes in isolation and on rerun.

## Fix
Inject the time source so the quiet period is virtual, not wall-clock:
- Parameterize the debounce delay behind an abstraction (`TimeProvider` on .NET 8+, or a `Func<TimeSpan, CancellationToken, Task>` delay delegate defaulting to `Task.Delay`).
- In tests, drive it with a controllable/`FakeTimeProvider` so batch boundaries are deterministic — no `Task.Delay(20)` race.
- Production `WatchSession` keeps the real `Task.Delay`; only the seam changes.

## Acceptance
The test passes deterministically 50/50 consecutive runs with no real sleeps in the assertion path; production watch behavior unchanged.

PR #734, opened 2026-07-22
test(watch): make every WatchDebouncerTests case deterministic (closes #729)

### LINKED ISSUE #729: Finish converting WatchDebouncerTests off wall-clock timing (residual flakes)

juanmicrosoft:
Closes #729.

## What

The residual `WatchDebouncerTests` flakes — sighted across the #723/#726/#728/#730 full-suite runs — came from cases that still measured the quiet period on the real `TimeProvider.System` clock (`Burst`, `SeparateBursts`) or a timed `CancellationTokenSource` (`Cancellation_WhileWaiting`). Under full-suite parallel load those real timers raced. Every case now drives a `FakeTimeProvider` on virtual time; **no test depends on the wall clock.**

## The subtle one

`EventWithinQuietPeriod` already used `FakeTimeProvider` but was still the flakiest. The debouncer parks on `WhenAny(quietTimer, moreEvents)`; the old driver advanced a **full quiet period on its first loop iteration**, firing the already-armed timer *before* the newly-written event was drained. `WhenAny` could then resolve the timer and close the batch **without** the event → `Assert.Contains("/p/b.calr")` fails.

The fix is a race-free driver (`DriveToClose`) with one invariant:

> **advance the clock only when the channel is empty.**

While an event is buffered the clock stays put, so the quiet timer *cannot* fire and `WhenAny` *must* resolve the event first — draining it into the batch and re-arming. The batch can never close with an un-drained event, no matter how starved the thread pool is. (Events are only written on the test thread, so `Count == 0` is a stable pre-advance check.)

## Verification

Root-caused by reproducing under load (the flake was invisible in isolation — the class passes 8/8 alone, but ~1-in-4 full-suite runs failed). After the fix:

- **10/10 full `Calor.Compiler.Tests` runs green** under parallel load (5865 passed).
- `WatchDebouncerTests` 7/7 in isolation (the class has 7 test methods).

No production code changed — the `WatchDebouncer`/`TimeProvider` seam from #714/#716 was already sufficient; this finishes converting the tests onto it.

### DIFF
--- a/tests/Calor.Compiler.Tests/WatchCommandTests.cs
+++ b/tests/Calor.Compiler.Tests/WatchCommandTests.cs
@@ -20,25 +20,29 @@ public class WatchDebouncerTests
 
     /// <summary>
     /// Drives <paramref name="time"/> forward until the debouncer's quiet timer fires
-    /// and the batch closes. Deterministic because it never advances the clock while
-    /// an event is still buffered: with the clock stationary the quiet timer cannot
-    /// fire, so the read always resolves a pending event (draining it and re-arming
-    /// the timer) rather than racing the timer against the event. Only advances a full
-    /// quiet window once the channel is empty. Bounded so a logic error can't hang.
+    /// and the batch closes.
+    ///
+    /// <para><b>Race-freedom invariant:</b> the clock is advanced ONLY when the channel
+    /// is empty. While an event is still buffered the clock stays put, so the quiet
+    /// timer cannot fire and <c>WhenAny</c> must resolve the event first — draining it
+    /// into the batch and re-arming. The batch therefore never closes with an un-drained
+    /// event, no matter how starved the thread pool is (<c>TryRead</c> is what empties
+    /// the channel, so <c>Count == 0</c> implies every written event is already batched).</para>
+    ///
+    /// <para><b>Precondition:</b> no events may be written to the channel after this
+    /// helper starts. Every write must happen before the drive; a mid-drive write from
+    /// another task would reintroduce the Count-check race. The loop is wall-clock
+    /// bounded so it keeps advancing as long as the runner is alive rather than
+    /// exhausting a fixed iteration budget and then parking on a timer no one fires.</para>
     /// </summary>
     private static async Task<IReadOnlyCollection<string>?> DriveToClose(
         Task<IReadOnlyCollection<string>?> read, FakeTimeProvider time, ChannelReader<string> reader)
     {
-        // Invariant that makes this race-free under any scheduling: advance the clock
-        // ONLY when the channel is empty. While an event is still buffered the clock
-        // stays put, so the quiet timer cannot fire and WhenAny must resolve the event
-        // (draining it into the batch and re-arming). The batch therefore never closes
-        // with an un-drained event, no matter how starved the thread pool is. The bound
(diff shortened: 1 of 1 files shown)


juanmicrosoft:
Thanks — all addressed, and thanks for tracing the interleavings independently.

**1. 8/8 → 7/7.** Fixed in the PR body. The class has 7 `[Fact]` methods (confirmed by `grep -c` and the runner); the "8" was a slip.

**2. Iteration bound → wall-clock deadline.** Swapped the fixed 1000-yield budget for `Stopwatch` vs `TestTimeout`:
```csharp
var deadline = Stopwatch.StartNew();
while (!read.IsCompleted && deadline.Elapsed < TestTimeout)
{
    await Task.Yield();
    if (reader.Count == 0) time.Advance(Quiet);
}
return await read.WaitAsync(TestTimeout);
```
It keeps advancing the *virtual* clock as long as the runner is alive, so it can't exhaust a budget and then park on a fake timer no one fires — the final `WaitAsync` is now a formality. The stopwatch is a liveness bound only; all timing semantics stay on virtual time.

**3. Precondition enforcement.** I kept it as a **bolded requirement** on the helper's doc rather than `Complete()`-ing the writer, because completing the channel changes what these tests verify: a completed channel makes the debouncer return the batch via the channel-completion path, *not* the quiet timer — which is exactly the behavior under test. So the open channel is load-bearing; the doc now states loudly that no write may occur after the drive starts.

**4. Notes.**
- Added the comment on `Cancellation_WhileWaiting` explaining why cancel-after-obtain is sound (synchronous run to the first await registers the token before `Cancel()`).
- Agreed on the duration-semantics division of labor — `EventWithinQuietPeriod…DurationIsHonored`'s partial-advance assertion is the only guard against an early-closing quiet window, and won't be deleted.
- `WatchSessionTests`' 30s eventually-completes bounds are noted as the next place to look if "watch tests flake" ever resurfaces — out of scope here.

Still 0/5 full-suite failures under parallel load after the change.

And agreed on the series wrap-up: this closes the last thread from the #716 review (TimeProvider seam, duration semantics, no tautological backstop, zero residual wall-clock tests).

Issue #705, opened 2026-07-16
Converter emits invalid Calor (broken indentation, use-before-def) for block-bodied lambdas passed as arguments


juanmicrosoft:
## Summary

`calor convert` (v0.7.0) produces **unparseable Calor** when the C# input contains a **block-bodied lambda** (statement lambda) passed as a method argument. The hoisted lambda's body statements are emitted at the wrong indentation level — dedented to near column 0 instead of being nested under the hoisted `§B{...}` binding — so the generated source fails to parse. Expression-bodied lambdas (`x => x * 2`) convert correctly.

Hit this converting the default ASP.NET Core minimal-API template (`Program.cs` with `app.MapGet("/weatherforecast", () => { ... })`), so any attempt to migrate a stock web project hits it immediately.

## Environment

- calor 0.7.0 (global dotnet tool), macOS (Darwin 25.5.0), .NET 10
- Reproduced via the `calor_convert` MCP tool; all conversion modes affected (`standard`, `passthroughOnError: true`, `conversionMode: interop`)

## Minimal repro

Input C#:

```csharp
public static class App
{
    public static void Main()
    {
        var items = new[] { "a", "b" };
        Handle("/x", () =>
        {
            var result = Enumerable.Range(1, 5).Select(index => items[index]).ToArray();
            return result;
        });
    }

    public static void Handle(string route, Func<string[]> handler) { }
}
```

Actual converter output (invalid — note the lambda body dedenting to column 2 inside a `§MT` whose statements are at column 6):

```
§M{m001:Repro}
  §CL{c001:App:pub:stat}
    §MT{m002:Main:pub:stat}
      §ARR{items:str}
        "a"
        "b"
      §B{_lam009} §LAM{lam004}
  §B{_lamSelect006} §LAM{lam005:index:?} §IDX{items} index §/LAM{lam005}
  §B{_chainRange007} §C{Enumerable.Range} §A 1 §A 5 §/C
  §B{_chainSelect008} §C{_chainRange007.Select} _lamSelect006
  §B{result} §C{_chainSelect008.ToArray}
  §R result
§/LAM{lam004}
      §C{Handle} §A "/x" §A _lam009 §/C

    §MT{m010:Handle:pub:stat} (str:route, Func<str[]>:handler)
      §R
```

Parse error:

```
Generated Calor failed to parse: Dedent to column 4 does not match any enclosing indent level.
```

[...]

PR #735, opened 2026-07-22
feat(convert): fall back to §CSHARP when emitted Calor does not parse (#717)

### LINKED ISSUE #717: Converter: fall back to §CSHARP when post-conversion validation fails (passthroughOnError)
Split from #705 (whose primary block-lambda indentation bug is now fixed). Defense-in-depth: when `calor convert`'s own post-parse validation detects the generated Calor does not parse, and `passthroughOnError`/`--no-fallback` semantics allow, it should wrap the offending member (or construct) in `§CSHARP{...}§/CSHARP` instead of writing invalid Calor.

Currently (#705 §'Fallback options don't mitigate') passthroughOnError produces identical broken output with interopBlocksEmitted:0. Acceptance: a construct that fails post-conversion parse validation is emitted as a §CSHARP interop block under the fallback modes, and the resulting file parses; add a conversion test where an intentionally-unsupported construct round-trips to a §CSHARP block.

juanmicrosoft:
Closes #717.

## What

The C#→Calor converter never re-parsed its own output. When an emitter bug produced malformed Calor (the motivating case was the #705 block-lambda indentation bug), `Convert` still returned `Success = true` with unparseable text, and `passthroughOnError` produced *identical* broken output — `interopBlocksEmitted: 0`.

`CSharpToCalorConverter.Convert` now adds a **post-emit validation step**: when the emitted Calor does not parse *and* a C#-preserving mode is active (`context.ShouldPreserveCSharp` — i.e. `passthroughOnError` or interop mode), each top-level member whose own emission does not parse is replaced with a `§CSHARP{…}§/CSHARP` interop block carrying its original C# (recovered from the Roslyn tree by kind/name), and the module is re-emitted. The result is **always valid Calor**; `interopBlocksEmitted` counts the fallbacks. Members that convert fine are left alone, and non-preserving modes are unchanged (no default-behavior change).

The `calor_convert` MCP tool already plumbs `passthroughOnError`, so it benefits directly.

## Why the §CSHARP block is safe for arbitrary C#

The lexer scans a `§CSHARP{ … }§/CSHARP` block to the literal `}§/CSHARP` terminator, so it round-trips arbitrary C# (braces, newlines) — the only thing it can't contain is the literal `}§/CSHARP` — a legal but vanishingly rare C# string literal.

## Testing — an honest note

This is genuine **deep defense-in-depth**. The visitor's own §CSHARP wrapping already handles unsupported features, and I **probed ~65 exotic C# constructs** (unsafe/pointers, `goto`, local functions, collection expressions, query syntax, raw strings, list patterns, ref structs, primary constructors, required members, conversion operators, `>>>`, function pointers, …) — **none reached this path**. That's strong evidence the converter is robust and this guards *future/unknown* emitter gaps, but it means a natural breaker isn't practically findable.

So the fallback is driven **deterministically** via an internal `ParseValidatorOverride` seam that declares a chosen member "unparseable", and then the final output is **parsed for real** to prove the fallback yields valid Calor. `PostValidationFallbackTests` covers:
- an "unparseable" member is wrapped in §CSHARP with its original C#, the whole file parses for real, and `interopBlocksEmitted == 1`;
- the neighbouring member still converts to a real Calor class (`§CL`), not wrapped;
- the fallback is **gated off** in non-preserving mode (no §CSHARP, `interopBlocksEmitted == 0`);
- a clean conversion triggers nothing.

## Scope / follow-up
- Gated on `ShouldPreserveCSharp` (the issue's `passthroughOnError` semantics); the CLI `convert` doesn't expose that mode today — exposing a `--passthrough`/interop flag for the CLI would be a small follow-up if wanted.
- Fallback granularity is top-level types (class/struct/record/interface/enum/delegate). A failing synthetic `Main` from top-level statements isn't rewrapped (no single C# source span) — a conservative miss.

## Verification
- `dotnet build` — 0 warnings.
- `Calor.Compiler.Tests` 5868 (incl. 3 new) / `Calor.Conversion.Tests` 286 — green.


juanmicrosoft:
Thanks — "the fallback trusts itself" is exactly right, and both boundary probes were real. Fixed the cluster.

**1. Cross-namespace collision → duplication (demonstrated).** `CollectTopLevelTypeSources` now groups sources per `kind/name` instead of blindly concatenating, and `TryTakeSource` recovers only the **safe** cases: exactly one declaration, or several that are **all `partial`** (a single merged type — concatenate). The ambiguous case — 2+ distinct same-named types in different namespaces — is **refused**, so a healthy `N2.Foo` is never dragged into `N1.Foo`'s §CSHARP block. Entries are also **removed on take**, closing the double-embed you noted. New test `CrossNamespaceSameNameCollision_DoesNotDuplicateTheHealthyType` pins it (no duplication; refused; surfaced as failure, not shipped).

**2. Silent give-up preserved the original bug (demonstrated).** The output is now **always re-validated** after the rewrap attempt. If it still doesn't parse, a warning `ConversionIssue` is added, and under `passthroughOnError` `Convert` returns **`Success = false`** — the passthrough contract is exactly "never hand me broken output." New test `UnrewrappableOutput_IsSurfaced_NotSilentlyShipped` uses the top-level-statements case (synthetic `Main`, no C# type source → genuinely unrewrappable) and asserts failure + warning + `interopBlocksEmitted == 0`.

**3. No post-rewrap re-validation.** That's the single `if` that now closes the loop (feeds finding 2). And you were right about the seam: I made the test override **honest** — it condemns the marker only when it appears **outside** a §CSHARP block, mirroring a real parser (a §CSHARP block's content is inert). So a member preserved inside §CSHARP legitimately passes re-validation, and the seam can break the first pass without lying about the final one.

**4. Smaller.**
- **Doc comments** are now preserved on rewrapped members — the XML doc trivia is prepended to the member body (which still uses `ToString()` to avoid namespace-indent bleed). Attributes already survived.
- **Module-level Functions** still aren't probed, but that's now an *acceptable* conservative miss because the give-up is loud (finding 2) rather than silent.
- **`}§/CSHARP` caveat** — agreed it was understated; it's a legal (if vanishingly rare) C# string literal. I'll soften the PR/doc wording accordingly.
- Filed **#736** for the CLI `--passthrough` flag so `calor convert` reaches the same protection the MCP tool already has.

Build 0 warnings; Compiler.Tests 5870 / Conversion.Tests 286 green.

Issue #733, opened 2026-07-22
calor -i accepts type-changing mutable §B rebind that emits CS0029


juanmicrosoft:
## `calor -i` accepts a type-changing mutable `§B` rebind that emits CS0029

Split from #727 (PR #730 review finding 3). A mutable rebind re-annotated with a different type is emitted as an assignment without a type check:

```
§B{~x:i32} 0
§B{~x:str} "hi"     // emits `int x = 0; x = "hi";` — CS0029
```

`calor -i` exits 0; `dotnet build` fails. The reassignment classification in `BindValidationPass` (`#727`) never compares the rebind's declared type to the original, even though the scope stack already carries declared types.

## Ask
- On a mutable rebind that carries an explicit type annotation, compare it to the variable's original declared type; if incompatible, emit a diagnostic (Calor02xx) at the rebind.
- The value-typed variant (`§B{~x} "hi"` when `x:i32`, no annotation) needs initializer type inference — may be a further follow-up.

## Acceptance
- The program above fails `calor -i`.
- Corpus stays clean (mutable rebinds re-using the same type are unaffected).

Issue #731, opened 2026-07-22
calor -i accepts same-scope duplicate §B that emits CS0128 (needs converter fix)


juanmicrosoft:
## `calor -i` accepts same-scope duplicate `§B` that emits CS0128

Split from #727 (PR #730 review finding 3). Two immutable bindings reusing a name in the **same** scope compile with exit 0 and emit invalid C#:

```
§B{x:i32} 1
§B{x:i32} 2      // emits `int x = 1; int x = 2;` — CS0128
```

`#727`'s `Calor0255` only flags shadowing of an *enclosing* scope (CS0136); same-scope duplicates (CS0128) are deliberately excluded because it interacts with the converter (below).

## Why it's separate
The C#→Calor converter emits array/list/dict **reassignments** as same-name creation blocks:
```csharp
int[] arr = new int[5];
arr = new int[] { 10, 20, 30 };   // converter emits a second §B{arr} array-creation block
```
Blanket-rejecting same-scope name reuse would flag this legitimate converter output. So closing CS0128 needs the converter to emit the reassignment as `§ASSIGN` (or a mutable rebind) rather than a fresh `§B` — a converter change, not just a pass check. (`ConverterQualityTests.Assignment_ArrayCreation_EmitsArrayBlock` and siblings exercise exactly this shape.)

## Ask
- Converter: emit `arr = …` reassignments as reassignment forms, not fresh `§B`.
- Pass: once the converter no longer produces spurious same-name `§B`, extend `Calor0255` (or a sibling code) to flag genuine same-scope duplicates.

## Acceptance
- The two-line program above fails `calor -i`.
- Converter round-trip tests stay green.

PR #737, opened 2026-07-22
fix(codegen): scope-aware mutable-rebind tracking to avoid CS0103 (#732)

### LINKED ISSUE #732: CSharpEmitter mis-compiles mutable §B rebind across sibling blocks (CS0103)
## CSharpEmitter mis-compiles a mutable `§B` rebind across sibling blocks → CS0103

Split from #727 (PR #730 review finding 2). The emitter's `_declaredVariablesInCurrentScope` is a **flat per-function** set, so a mutable `§B{~x}` in a sibling block is emitted as an assignment to an out-of-scope variable:

```
§IF{i1} (== a true)
  §B{~x:i32} 1
§IF{i2} (== b true)
  §B{~x:i32} 2      // emits `x = 2;` — but x from the first block is out of scope → CS0103
```

`calor -i` exits 0; `dotnet build` fails. The correct emission is `int x = 2;` (a new declaration in the sibling block, which is valid — sibling scopes).

## Root cause
`CSharpEmitter._declaredVariablesInCurrentScope` never removes names when a block closes, so it thinks `x` is still declared in the second block and emits a reassignment.

`BindValidationPass` (`#727`) intentionally mirrors this flat set (`_declaredInFunction`) to classify mutable rebinds, so it does **not** flag this — it faithfully certifies the emitter's bug. The PR #730 review's point: the pass should agree with Roslyn, not the emitter.

## Ask
- Make `CSharpEmitter`'s declared-variable tracking **scope-aware** (push/pop per block), so a rebind in a sibling block re-declares instead of reassigning.
- Make `BindValidationPass._declaredInFunction` scope-aware to match, so its reassignment classification is correct.
- The differential Roslyn test added in #730 (`ShadowingDifferentialTests`) has this case marked as a known gap referencing this issue — flip it to the clean assertion once fixed.

## Acceptance
- The program above compiles to valid, CS0103-free C# (or fails `calor -i` with a diagnostic).
- Snapshot/golden codegen tests stay green (only sibling-mutable-rebind output changes).

juanmicrosoft:
Closes #732. The last emitter-level bug in the shadowing/rebind family.

## The bug

`CSharpEmitter` tracked declared locals in a **flat per-function** `HashSet`, so a mutable `§B{~x}` in a sibling block was emitted as an assignment to a variable that no longer exists:

```
§IF{i1} (== a true)
  §B{~x:i32} 1        // int x = 1;
§IF{i2} (== b true)
  §B{~x:i32} 2        // was: `x = 2;` → CS0103 (x from the first block is out of scope)
```

`calor -i` exited 0; `dotnet build` failed with CS0103. The correct emission is a second `int x = 2;` — valid, because sibling scopes are independent.

## The fix

The emitter now tracks declarations in a **scope stack** (`_declScopes`): reset to a single base scope at each function/method/accessor/ctor, **pushed on entering a control-flow block** (if/elseif/else, while, do-while, for, foreach, dict-foreach, match case, try/catch/finally, using, sync, unsafe) and **popped on exit**. A mutable rebind emits a reassignment only when the name is visible in a live scope:

| Case | Before | After |
|---|---|---|
| Sibling rebind (closed sibling) | `x = 2;` → **CS0103** | `int x = 2;` ✓ |
| Accumulator (`§B{~result} (* result i)` in a loop, outer `result` live) | `result = result * i;` | `result = result * i;` (unchanged) |

`BindValidationPass`'s reassignment classification was made scope-aware to match — the flat `_declaredInFunction` mirror is replaced by `IsDeclaredInAnyLiveScope` over the scope stack #727 already maintains, so the pass agrees with the fixed emitter (per the #730 review's "agree with Roslyn, not the emitter").

## Tests
- **`ShadowingDifferentialTests`**: the sibling-mutable-rebind case moves from a `KnownGap_*` (asserting CS0103) into `CleanWhenAccepted` — the differential invariant (*accepted ⟹ Roslyn-clean*) now holds for it.
- **`SiblingRebindCodegenTests`** (new): asserts two independent `int x` declarations for the sibling case, and that the accumulator stays a single declaration + reassignment.

## Verification
- `dotnet build` 0 warnings; `calor self-test` **10/10** goldens unchanged (the broad emitter change altered only sibling-mutable-rebind output).
- Compiler.Tests 5872 / Conversion 286 / Evaluation 199 / Semantics 37 / Enforcement 271 — all green.
- `self-check docs` clean.

Remaining shadowing/rebind gaps: **#733** (type-changing rebind, CS0029) and **#731** (same-scope duplicate, CS0128) — both still pinned in `ShadowingDifferentialTests`.


juanmicrosoft:
Thanks — you were exactly right: the fix aligned the two scope models for §B locals but not for construct-introduced variables. Fixed the whole class.

**1. Loop variables (blocking, demonstrated).** The emitter now registers each construct's variables right after `PushDeclScope()` — `§L` loop var, `§EACH`/`§EACHKV` iteration vars, dict-foreach key/value, `catch` exception var, `using` binding. So `§L{l1:i} … §B{~i} 9` now emits `i = 9;` (valid — for-loop vars are reassignable), matching the pass. Verified via the differential invariant: accepted **and** Roslyn-clean.

**4. Parameters (same class).** Also fixed — `ResetDeclScopes` now seeds parameters into the base scope at function/method/ctor/operator entry, so `§B{~param} 9` emits `param = 9;` instead of a CS0136 re-declaration. (The pass already held parameters in its scope 0, so they now agree.)

**2. Differential coverage — the durable part.** Added the construct-variable rebind rows you called for:
- `for-loop-var-rebind` and `parameter-rebind` → `CleanWhenAccepted` (now valid).
- `foreach-variable-rebind` → a new `KnownGap_*` asserting **CS1656**. This one is genuinely different: a `foreach` iteration variable is *not assignable* in C#, so once both models agree it's in scope, the reassignment itself is invalid — there is no correct emission. The only right answer is a reject diagnostic, filed as **#738**; the row pins CS1656 until then. (This is the honest resolution of "for/foreach lumped together" — for-loop vars are reassignable and fixed; foreach vars are not and need rejection.)

These rows are the mechanism you named — they'd have caught finding 1 before I shipped it.

**4b. Balance asserts.** `PopDeclScope`/`DeclareVarInScope` now `Debug.Assert` scope balance instead of silently masking it. The assert *immediately* earned its keep — it caught isolated statement/expression emission paths (many unit tests) that never establish a base scope; fixed by initializing `_declScopes` with a base scope, so the guard is a real invariant now, not a silent fallback.

**3. Two hand-maintained scope stacks.** Agreed this is the structural risk, and agreed it's not a merge blocker. I'm treating the differential rows as the safety net for now (any future construct-variable divergence fails a test); a shared scope-tracker consumed by both components is the right eventual refactor and I'll note it for the backlog.

self-test 10/10 goldens unchanged; Compiler.Tests 5875 / Conversion 286 / Evaluation 199 / Semantics 37 / Enforcement 271 / Verification 253 — all green.

Issue #738, opened 2026-07-22
§B rebinding a §EACH/§EACHKV iteration variable emits CS1656 (needs a reject diagnostic)


juanmicrosoft:
## `§B{~x}` rebinding a `§EACH`/`§EACHKV` iteration variable emits CS1656

Surfaced during the PR #737 review (finding 1). After #732 made the emitter's scope model register construct-introduced variables, a mutable `§B` that rebinds a **foreach iteration variable** is classified as a reassignment by both the pass and the emitter — but a C# `foreach` iteration variable is **not assignable**:

```
§EACH{e1:x} arr
  §B{~x:str} "y"     // emits `x = "y";` inside `foreach (var x in arr)` → CS1656
```

`calor -i` exits 0; `dotnet build` fails with CS1656 ("Cannot assign to 'x' because it is a 'foreach iteration variable'"). Same for `§EACHKV` key/value variables (tuple-deconstruction iteration variables, also non-assignable), and the `foreach (var (k,v) in …)` dict form.

## Why it's separate from #732
`§L` for-loop variables and parameters **are** reassignable, so #732 correctly registers them and the emitter emits a valid `i = 9;`/`param = …;`. Foreach iteration variables are the exception: neither a reassignment (CS1656) nor a re-declaration (CS0136) is valid, so there is no correct emission — the only right answer is to **reject** the rebind with a diagnostic.

## Ask
- Add a semantic diagnostic (Calor02xx) rejecting a `§B` whose name is a `§EACH`/`§EACHKV` iteration variable in scope (the iteration variable is read-only for assignment purposes).
- `BindValidationPass` already registers these variables in its loop-var scope; distinguish "iteration variable" from a normal local so it can reject the rebind rather than silently classify it as a reassignment.

## Acceptance
- The program above fails `calor -i` with a diagnostic pointing at the `§B`.
- `ShadowingDifferentialTests` foreach/each-kv rebind rows move from KnownGap to rejected.

## Context
- PR #737 pins these as `KnownGap_*` rows asserting CS1656, referencing this issue.

PR #739, opened 2026-07-22
feat(binder): reject type-changing mutable §B rebind (Calor0256) (#733)

### LINKED ISSUE #733: calor -i accepts type-changing mutable §B rebind that emits CS0029

juanmicrosoft:
Closes #733.

## What

A mutable `§B` rebind re-annotated with a different type was accepted, then emitted broken C#:

```
§B{~x:i32} 0
§B{~x:str} "hi"     // emitter: `int x = 0; x = "hi";` → CS0029
```

`calor -i` exited 0; `dotnet build` failed. `BindValidationPass` classified the second `§B` as a reassignment (mutable, `x` in scope) but never compared the rebind's declared type to the variable's original.

## The fix

On a mutable reassignment carrying an explicit annotation, the rebind type is compared to the variable's declared type (from the scope stack #727/#732 already maintains); a mismatch reports **`Calor0256`** at the rebind. The variable's type is fixed at its first declaration.

**The one subtlety** — and the false positive I caught via the differential suite: `§B` annotations are stored already-expanded (`INT`) while parameter/loop-variable types are raw (`i32`), so a naive string compare flags a *matching* parameter rebind (`§F(i32:x)` then `§B{~x:i32} 9`). Both sides are now canonicalized through `AttributeHelper.ExpandType` before comparing, so `i32 ≡ INT` and only genuine mismatches fire.

**Unaffected** (verified): same-type rebinds; the unannotated accumulator idiom (`§B{~result} (* result i)` — no annotation to contradict); an unannotated/inferred original (no known type); and sibling same-named variables of different types (those are new declarations, not rebinds).

## Tests
- `ShadowingDifferentialTests`: the type-changing-rebind case moves from `KnownGap_*` (CS0029) into `RejectedIdioms` — the differential invariant now covers it.
- `RebindTypeMismatchTests` (new): reject (different type), accept (same type), accept (unannotated rebind), accept (untyped original), accept (sibling different types).
- `BindCorpusCleanTests` pins `Calor0256` clean across `samples/` + benchmarks.

## Verification
- `dotnet build` 0 warnings; `calor self-test` 10/10; `self-check docs` clean.
- Compiler.Tests 5880 / Conversion 286 / Evaluation 199 / Semantics 37 / Enforcement 271 — all green.

Remaining shadowing/rebind gaps: **#731** (same-scope duplicate, CS0128 — needs converter cooperation) and **#738** (foreach-var rebind, CS1656) — both still pinned in `ShadowingDifferentialTests`.


juanmicrosoft:
Thanks — both addressed, and the systemic one folded in with a helper the whole series can use.

**1. Unannotated variant (demonstrated).** Fixed the common case: `Calor0256` now takes the rebind's type from the explicit annotation when present, **otherwise from a literal initializer** whose type is statically known (int/str/bool/float). So `§B{~x:i32} 0` then `§B{~x} "hi"` is now rejected too. The remaining lane — an unannotated **non-literal** mismatched value (needs value-type inference) — is filed as **#740** and pinned as a new `KnownGap_UnannotatedNonLiteralRebind_EmitsCS0029` row (accepts, CS0029), so the suite enumerates it rather than hiding it. CHANGELOG now carries the scope sentence.

**2. Internal type spellings (recurring).** Fixed point-wise **and** folded in the durable helper: added `AttributeHelper.ToSurfaceSpelling` — the inverse of `ExpandType`, recursive over arrays/options/results/generics — and the `Calor0256` message now reads `declared 'i32' but is rebound as 'str'`. A test asserts the message contains `i32`/`str` and **not** `INT`/`STRING`, and `ToSurfaceSpellingTests` round-trips `ExpandType` including composites (`ARRAY[element=STRING]` → `[str]`, `List<STRING>` → `List<str>`). You're right this is finding-N in the series, so I filed **#741** for the systemic finish: route every type-echoing diagnostic through `ToSurfaceSpelling` and add a corpus/test-idiom guard asserting no internal spelling leaks — so the next new diagnostic that regresses fails a test, not a review.

**3. Noted.** Added the one-line comment on why `TryLookupLocal`'s field fallback is unreachable in this path (the reassignment guard requires the name in a live scope). Repeated-mismatch cascading is intentional (each rebind reported once, variable keeps its original type). Remaining-gaps bookkeeping updated: **#740** (non-literal unannotated), **#738** (foreach CS1656), **#731** (same-scope CS0128) — all pinned in `ShadowingDifferentialTests`.

self-test 10/10; corpus clean; Compiler.Tests 5903 / Conversion 286 / Semantics 37 / Enforcement 271 — all green.

PR #757, opened 2026-07-24
feat: final envelope sweep — M-E1 reaches 100% (loop plan D1.3, WS1 exit)

juanmicrosoft:
Completes the WS1 adoption sweep started in #754: the last 8 denominator surfaces now emit envelope schema v1.1, bringing **M-E1 to 29/29 = 100%** — the WS1 exit criterion from `docs/plans/loop-plan-v0.9.md`.

## Surfaces
- **migrate CLI** — the `--report *.json` file is now envelope-wrapped (`command: "migrate"`, report unchanged under `data`); per-file conversion issues surface as `Calor1343` diagnostics. Also fixes migrate's `Environment.ExitCode` stomp (same bug class as #754's verify/convert fix — error paths actually exited 0).
- **calor_convert** — conversion issues become `Calor1343` envelope entries (`[feature]`-prefixed); validate-mode diagnostics are real parse/compile diagnostics with `declarationId`.
- **calor_batch** — per-file error strings become envelope entries (real compiler diagnostics in compile mode); `errorCategories`/summary kept.
- **calor_migrate** — per-file `"[Code] Ln: msg"` strings become envelope entries in all phases; `errorCategories` keyed by diagnostic code.
- **calor_navigate / calor_structure** — parse-error strings become envelope entries via a shared `ParseResult.ToEnvelopeDiagnostics()` (EditPreviewTool refactored onto the same helper).
- **calor_format** — format errors are real parser diagnostics; ids-action issues are `Calor0800`-band diagnostics with `declarationId` (mirrors the CLI `ids check` adoption).
- **calor_fix** — adopted **by audit**: its payload is applied-fix records only; verified no diagnostic-shaped data hides in it (error paths use protocol-level MCP errors). No shape change.

The denominator table in `docs/cli/envelope-schema.md` has zero `No` rows left and now states the 100% figure plus the forward rule: new commands/tools must be added to the table (adoption or reviewed exemption) before shipping.

## Breaking (v0.9 pre-release, same envelope unification as #754)
MCP consumers reading the old flat shapes must switch to envelope entries: `calor_convert`/`calor_batch` `issues[]`, `calor_migrate` `errors[]`/`warnings[]`, `calor_navigate`/`calor_structure` `errors[]`, `calor_format` `issues[]`. CHANGELOG has the migration notes.

## Testing
[...]

PR #758, opened 2026-07-24
feat: M2 — loop telemetry v2, baseline archive, Call 1 record (loop plan WS4)

juanmicrosoft:
Executes milestone **M2** of the v0.9 loop plan (#747): the measurement infrastructure that must exist before any WS2/WS3 loop-improvement code merges.

## D4.3a — baseline archived (the plan's hard rule)
Annotated tag **`loop-baseline-ws1`** → `4f235cdc` (post-#757, WS1-only build), pushed to origin with pins (v0.8.0, SDK 10.0.100, Z3 4.15.7, Roslyn 5.3.0). This is M5's control arm A, to be checked out and run **simultaneously** with the treatment build — never compared longitudinally.

## D4.1 — loop telemetry schema v2
`bench/phase0-agent-native/loop-telemetry-schema.md` + machine-validatable `loop-telemetry.schema.json`. Additive over the existing `journal.jsonl` shim stream: `feedback_latency_ms`, `iteration` (gates-doc §2 semantics — edited invocations only), `edit_mechanism`, `edit_target_ids` (raw-file node attribution so M-L3's denominator doesn't narrow to MCP edits), envelope `diagnostics` `{code, declarationId}` + `envelope_valid`, `src_tree_hash` as snapshot reference, and reserved-null `apply_verdict`/`rejected_edit` so baseline and treatment arms emit identical shapes — the schema freezes before the treatment build exists.

## D4.2 — harness emission + arm-constraint capability
`run-pair.sh`'s `dotnet` shim now emits v2 records (portable ms clock, per-run iteration counter, opt-in `CALOR_LOOP_SNAPSHOTS=1` tree archiving, envelope capture on source copies so `.g.cs` never pollutes the agent workspace or edit-detection hash). New `--edit-mechanism raw|mcp-file|mcp-node` arm constraint following the `--exemplar` label pattern with PreToolUse-hook enforcement (`mcp-node` accepted-with-warning: descoped per Call 1). `run-epoch.sh` pins gain `editMechanism` + `telemetrySchema`; `result.json` gains `meanFeedbackLatencyMs`/`envelopeValidAll`. Includes `validate-telemetry.sh` (jsonschema or stdlib fallback) and `telemetry-helpers.py`. Also fixes a pre-existing jq 1.7 bug that aborted every epoch rollup.

**Verified spend-free**: null-agent runs on both arms (zero API calls) — journals 4/4 schema-valid, iteration math identical to v1 semantics, `envelope_valid: true` on a clean build, and `{"code":"Calor0410","declarationId":"f001"}` extracted from a seeded effect violation. No real `claude` invocations anywhere.

## Call 1 record + D4.5 dry-run design
`docs/plans/loop-m2-baseline.md`:
- Records the **Call 1 input**: machine-zone §13's E1 verdict (H1 KILLED, 55% pooled) triggers the plan's **pre-committed** scope gate — WS2 descopes to D2.4 (file-level transactional apply) + D2.5 (write-path robustness); PP-L3 retired without running.
- **D4.5 feasibility dry-run design** (`loop-feasibility-dry-002`): variance estimate on the baseline build with pre-stated decision rules — including the sub-integer detectability trap (15% of a median-3 iterations-to-green is 0.45) and the tokens-to-green fallback — plus the spend-numbers template. **No epoch runs until the numbers are entered and authorized.**

## Also
- Removes accidental build artifacts from `tests/TestData/Verification/Outcomes/` (owed from #757) and gitignores the pattern.
- Known limits stated in-doc: `feedback_latency_ms` excludes the shim's post-stamp bookkeeping (identical across arms); envelope capture mirrors the arm build's enforcement (no `--verify`); no pair currently has literal `gateEligible: true` (verification used a non-retired pair).

No `src/` changes — `calor-first-guard` unaffected; compiler test suites unaffected.

Remaining for M2 completion after this merges: run the calibration + authorize spend for `loop-feasibility-dry-002`, run the threshold epoch on the baseline, freeze `[P]` thresholds via the gates-doc annex (D4.4).


juanmicrosoft:
Review addressed — both blockers and the spend guardrail, verified spend-free (null-agent only, no claude invocations):

**1 (🔴 mcp-file fabricates failure data):** gated. A live `--edit-mechanism mcp-file` run now exits 2 with an explanatory error (no MCP server registered, no file-writing MCP tool exists — verified); `--null-agent` keeps the labels/pins/record plumbing testable. Unblocks when WS2 D2.4 ships an actual MCP write path. The M2 doc's Call 1 paragraph is corrected from 'kept cheaply' to the gated reality.

**2 (🔴 latency arm-bias):** the stamp moved to the moment the agent-visible `dotnet` returns — before the silent src rebuild and held-out suite. Verified on null-agent runs: calor-arm build latency drops ~3.6s → ~1.1s, i.e. the arm-asymmetric held-out cost is out of the metric. The schema doc's latency bullet now states these exact semantics and the fairness rationale.

**3 (🟠 watchdog spend hole):** the harness now prefers coreutils `timeout`/`gtimeout` (kills the claude process directly, `-k 10` grace); the bash fallback enables job control so the backgrounded subshell is a process-group leader and `kill -9 -- -pid` reaches the reparented child. (Group-kill semantics are the standard fix; the live-claude path itself remains untested by design — no API calls in verification.)

**Minor:** schema code pattern widened to compiler-style codes so helper and schema agree (`CS####` legal, doc notes it); the shim's telemetry section is serialized with a portable mkdir lock (bounded spin, degrades to pre-lock behavior rather than deadlocking) so concurrent invocations can't corrupt iteration ordinals. On the 'verified spend-free is opt-in' phrasing: correct — the guarantee is scoped to the verification methodology (`--null-agent`), not the default run path, which exists to spend by design under the authorization gate.

Post-fix verification: journals still 4/4 schema-valid, iteration math unchanged, gate exit codes confirmed.

PR #795, opened 2026-07-24
docs: instrument-metrics Annex A in gates doc — D4.4 freeze, Call 1 closed

juanmicrosoft:
Adds **Annex A** to `docs/plans/agent-native-gates.md` per loop plan D4.4: additive-only, observational-only, own version counter (A-1.0), no gate-criterion cross-references — the main document's §1–§7 are untouched.

Frozen from the `loop-feasibility-dry-002` variance epoch, before any treatment build exists:
- **PP-L5**: ≥15% relative reduction in median per-pair **tokens-to-green** (A vs B simultaneous, ≥7 pairs × ≥5 runs/arm). Iterations-to-green demoted to observational — the dry-run showed it floor-bound (median 1, 94% at floor), so the pre-registered fallback applied before freezing, as D4.5 requires.
- **PP-L4**: reported-not-adjudicated (3 qualifying M-L3 events vs the 20-event floor).
- **PP-L3**: retired unrun (machine-zone E1 killed H1).
- **PP-L1/PP-L2**: unchanged / already met at WS1 exit.

Also marks **Call 1 closed** in the M2 record. This completes M2; next milestone is M3 (descoped WS2). Maintainer approved the tokens threshold 2026-07-24.


juanmicrosoft:
Review addressed:

**1 (🟠 pre-registration authority not on main) — accepted; PR gated.** Converted to **draft** until #747 (the loop plan) merges — the reviewer is right that a pre-registration freeze whose authority document is a mutable open draft is unverifiable from the repo, and the ordering was backwards. The annex header now states the merge-order dependency explicitly. @juanrivera: #747 is your review to complete — once it merges (with the §4/§5/D4.4/D4.5 rules as this annex cites them), I'll flip this back to ready. If #747's review *changes* any cited rule, this annex must be re-derived, not merged.

**2 (🟡 'observational-only' contradiction) — reworded.** The header now states precisely what holds: no §1–§7 machine-adjudicable **language-gate** criterion references annex metrics, while the annex's own A.2 thresholds ARE adjudicated pass/fail criteria — for the loop program's tooling-investment decisions, a separate question from the C#-vs-Calor gates. 'Observational-only' is no longer overclaimed.

**3 (🟡 7-cluster MDE fragility) — caveat carried into A.2.** The PP-L5 row now discloses the design-stage simulation basis (7 clusters, 400×200 bootstrap), the wide uncertainty on the MDE itself, and the consequence: if M5's realized variance is materially higher, a miss is reported as underpowered rather than adjudicated as a clean miss.

**Minor (self-asserted approval)** — now disclosed in the A.3 revision log per the main document's §7 bus-factor-1 convention.

PR #797, opened 2026-07-24
feat: M3 — MCP project sessions + transactional file write (loop plan WS2 D2.1/D2.4)

juanmicrosoft:
## Summary

M3 implementation slice 1 per the kickoff record (`docs/plans/loop-m3-ws2.md`, #796):

- **D2.1 Project sessions** — `calor_session_open` / `calor_session_close`: session-scoped project context in the MCP server. Opens a directory, parses every `.calr` file under it, holds state in memory. Dirty-state invalidation is stat-on-access (mtime/size gate, then SHA-256 content hash — the `BuildStateCache` two-level pattern) and picks up files changed, added, or deleted behind the session's back. No project-file format introduced (priced decision, kickoff §2).
- **D2.4 Transactional whole-file apply** — `calor_file_write`: runs the `calor_edit_preview` check set (compile / contracts / effects / references) against the current on-disk content, applies atomically (temp file + rename in the target directory) on `safe`/`safe_with_warnings`, rejects `breaking` with envelope schema v1.1 diagnostics and the file untouched. With a `sessionId`, reference checks widen to the whole project at identical strictness (same symbol-containment heuristic as the in-file check).
- **D2.5, auto-heal half** — `SourceHealer` wired into the write path: forbidden closers / indentation slips are healed before checking (default on, `heal: false` to disable); the healed text is what gets written, returned as `writtenContent` with ambiguities as `healNotes`. The fault-tolerant parse mode half of D2.5 is PR 3.
- Check core of `EditPreviewTool` widened `private → internal` so both tools share one check set (D2.4's "identical checking" requirement) — no logic changes.
- Envelope denominator updated per the WS1 forward rule: `calor_session_open` (E), `calor_file_write` (E), `calor_session_close` (X, lifecycle-only); M-E1 = 31/31.

## Dependencies

Depends on #796 (calor-first allowlist entries for the four new C# files). **calor-first-guard will fail here until #796 merges and this branch is rebased onto main** — the guard requires allowlist entries on main at the merge base and a PR cannot self-exempt.

## Test plan

- 17 new tests in `SessionAndFileWriteToolTests`: session open/parse-error/close lifecycle, apply/reject transactionality (file untouched on `breaking`), file creation, heal-then-apply vs `heal:false` reject, unknown-session and outside-root refusals, cross-file dangling-reference rejection, behind-the-back change invalidation, post-apply session-state freshness, handler registration.
- Full suite green locally: 7,600+ tests across all projects, 0 failures (`TreatWarningsAsErrors` on).

### DIFF
--- a/docs/cli/envelope-schema.md
+++ b/docs/cli/envelope-schema.md
@@ -171,11 +171,15 @@ Classes:
 | `calor_structure` | E | **Yes** | parse errors as envelope entries |
 | `calor_format` | E | **Yes** | parser + `Calor0800`-band diagnostics as envelope entries with `declarationId` |
 | `calor_fix` | D | **Yes** (by audit) | applied-fix records only; verified no diagnostic-shaped data hides in the payload |
+| `calor_session_open` | E | **Yes** | per-file parse errors as envelope entries under `diagnostics[]` (loop plan WS2 D2.1) |
+| `calor_session_close` | X | — | session lifecycle only; no source-anchored diagnostics |
+| `calor_file_write` | E | **Yes** | check-set errors as envelope entries (`compilationResult.errors`, same shape as `calor_edit_preview`); apply verdict + heal payload alongside (loop plan WS2 D2.4/D2.5) |
 | `calor_help` | X | — | documentation lookup; no source-anchored diagnostics |
 | `calor_self_test` | X | — | golden-diff scenarios |
 
 **M-E1 (envelope coverage)** = adopted E+D surfaces / all E+D surfaces =
-**29/29 = 100 %** as of the final D1.3 sweep — the WS1 exit criterion is met.
+**31/31 = 100 %** (29 at the final D1.3 sweep, +`calor_session_open` and
+`calor_file_write` from loop plan M3/WS2) — the WS1 exit criterion is met.
 Any new command or MCP tool must be added to this table (with an envelope
 adoption or a reviewed exemption) before it ships; the conformance suite and
 this table are the drift guards.
--- a/docs/plans/loop-m3-ws2.md
+++ b/docs/plans/loop-m3-ws2.md
@@ -33,11 +33,23 @@ format would be unpriced scope. Revisit trigger: WS3 D3.3 (10k-line fixture)
 or v0.10 multi-project needs.
 
 **Dirty-state invalidation is stat-on-access, not a watcher.** Each tool call
-that touches session state re-stats its files (mtime/size gate, then SHA-256
-content hash — the `BuildStateCache.IsFileUpToDate` two-level pattern) and
-reparses only changed files. A filesystem watcher adds lifecycle complexity a
-single-client stdio server does not need. Revisit trigger: WS3 D3.1 warm-state
-latency work, where re-stat cost appears in M-L1.
+that touches session state re-stats its files (mtime/size gate; SHA-256 hash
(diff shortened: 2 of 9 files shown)


juanmicrosoft:
Review response — all findings addressed in 30af954c. Point-by-point:

**🔴 B1 (no-session arbitrary write) — fixed.** Without a session, writes are now confined under the server's working directory (captured at tool construction); with one, under the session root. Session roots are themselves confined under the server root, since opening a session confers write access — so there is no session-shaped hole either. New tests: `FileWrite_NoSession_OutsideWriteRoot_ReturnsError`, `SessionOpen_OutsideServerRoot_ReturnsError`.

**🔴 B2 (symlink bypass) — fixed.** New `CanonicalPath` resolver walks every path component, resolving link chains (depth-capped) including links whose targets contain further links; containment compares canonical paths on both sides. The `.calr` suffix is also checked post-resolution. New test: `FileWrite_SymlinkedSubdirectoryEscape_ReturnsError` (lexically-inside path through a symlinked subdir → rejected, nothing written outside).

**🟠 M2 (unsound reference check) — fixed in both directions.** Removed functions are now matched against actual call targets collected from each file's AST via `CallGraphAnalysis` (calls reference *names*, e.g. `§C{add}`), not id-substrings. Types use whole-word matching; files that don't parse fall back to whole-word text so a broken neighbor can't fully hide a reference. Applied to `calor_edit_preview`'s in-file check and the project-wide check identically. The old contrived test (string literal `"wraps f001"` expected to block) is now inverted: `FileWrite_StringMentionOfRemovedFunction_DoesNotBlock` plus `FileWrite_RemovingCalledFunction_RejectsAsBreaking` with a real call site.

**🟠 M3 (TOCTOU) — mitigated to the extent possible without OS file locks.** A per-canonical-path semaphore serializes check→apply across concurrent tool calls, and on-disk content is revalidated against the checked original immediately before the rename (mismatch → error, retry). The residual race — an *external* writer between revalidation and rename — is documented in the tool header and kickoff doc as accepted for a single-client stdio server.

**🟠 M1 (stat-gate description) — claim corrected rather than hashing every file every call.** The gate now documents BuildStateCache semantics honestly: mtime+size-preserving edits are undetected until the stat changes; the hash exists to suppress reparses of touched-but-unchanged files. Comments, class docs, and the kickoff doc all state the limitation; the +5s bump in the behind-the-back test is annotated as granularity determinism, not a workaround.

**🟡 Minors — all addressed:** permissions preserved on rewrite (`FileWrite_PreservesUnixFileMode`); session open streams enumeration under the 2000-file cap, skips reparse points (no symlink cycles or out-of-root pull-ins), refuses >512 KB files as parse-error entries (`SessionOpen_OversizeFile_ReportedAsParseError`); all three tools wrap their flows so IO races / invalid paths surface as tool errors, not protocol crashes; heal now **caps the verdict at `safe_with_warnings` whenever `healApplied`** (heal test asserts it). FIFO-shaped inputs remain a residual (no portable non-blocking probe in .NET) — noted, root-confined.

**One pushback:** the "dangling authority" note is stale — plan #747 merged to main as `762adb0b` (and the Call 1 record in #758/`b6bd901e`), so the kickoff doc's §-references resolve on main today.

Full suite re-run after the fixes: 7,600+ tests, 0 failures, `TreatWarningsAsErrors` on.

PR #798, opened 2026-07-25
feat: M3 — fault-tolerant parse mode for the MCP write path (loop plan WS2 D2.5)

juanmicrosoft:
## Summary

Completes D2.5 (the auto-heal half shipped in #797). Per the kickoff record (`docs/plans/loop-m3-ws2.md` §3, PR 3): a first-class fault-tolerant parse mode surfaced through the write path's reject envelope.

- **`CalorSourceHelper.ParseTolerant`** — returns the parser's best-effort AST alongside diagnostics instead of discarding the tree on the first error (`ParseResult.Partial` / `IsPartial`). Lexer errors no longer abort the parse; an unrecoverable parser throw degrades to a plain failure rather than propagating. A partial result is **never** a compile pass — all verdict logic keys off `IsSuccess`, unchanged; a partial AST cannot upgrade a breaking edit.
- **Declaration-id attribution on rejects** — parse errors in `calor_file_write` rejects and `calor_session_open` diagnostics now carry `declarationId`, resolved by span containment against the partial AST (`DeclarationIdResolver`): the agent learns *which* function is broken, not just a line number.
- **Broken neighbors keep their surface** — session files parse tolerantly, so a file with an error in one function still contributes real call targets from its intact functions to the project-wide reference check: a genuine `§C{add}` there vetoes removing `add` (labeled "file has parse errors"), while a string-literal mention still does not. Whole-word text fallback now applies only to files with no AST at all.
- Edit summaries survive a broken side (symbol diff runs on partial ASTs).
- Denominator notes updated for the `declarationId` enrichment; no shape changes.

## Test plan

- 9 new tests in `TolerantParseTests`: partial-AST shape (intact declaration survives), declarationId attribution (unit, write-path reject, session-open), broken-neighbor real-call veto and string-mention non-veto, garbage/empty inputs never throw.
- Full suite green locally: 7,600+ tests, 0 failures, warnings-as-errors, Release binary rebuilt.

Remaining for M3: PR 4 — harness MCP registration + `mcp-file` gate unblock + WS2 exit measurement (M-L2/M-L4).

### DIFF
--- a/docs/cli/envelope-schema.md
+++ b/docs/cli/envelope-schema.md
@@ -171,9 +171,9 @@ Classes:
 | `calor_structure` | E | **Yes** | parse errors as envelope entries |
 | `calor_format` | E | **Yes** | parser + `Calor0800`-band diagnostics as envelope entries with `declarationId` |
 | `calor_fix` | D | **Yes** (by audit) | applied-fix records only; verified no diagnostic-shaped data hides in the payload |
-| `calor_session_open` | E | **Yes** | per-file parse errors as envelope entries under `diagnostics[]` (loop plan WS2 D2.1) |
+| `calor_session_open` | E | **Yes** | per-file parse errors as envelope entries under `diagnostics[]`, with `declarationId` attributed via tolerant-parse partial ASTs (loop plan WS2 D2.1/D2.5) |
 | `calor_session_close` | X | — | session lifecycle only; no source-anchored diagnostics |
-| `calor_file_write` | E | **Yes** | check-set errors as envelope entries (`compilationResult.errors`, same shape as `calor_edit_preview`); apply verdict + heal payload alongside (loop plan WS2 D2.4/D2.5) |
+| `calor_file_write` | E | **Yes** | check-set errors as envelope entries (`compilationResult.errors`, same shape as `calor_edit_preview`), with `declarationId` on parse errors via tolerant-parse partial ASTs; apply verdict + heal payload alongside (loop plan WS2 D2.4/D2.5) |
 | `calor_help` | X | — | documentation lookup; no source-anchored diagnostics |
 | `calor_self_test` | X | — | golden-diff scenarios |
 
--- a/src/Calor.Compiler/Diagnostics/Diagnostic.cs
+++ b/src/Calor.Compiler/Diagnostics/Diagnostic.cs
@@ -87,6 +87,7 @@ public static class DiagnosticCode
     /// the column of the nearest preceding <c>§IF</c>.
     /// </summary>
     public const string MisalignedElseClause = "Calor0117";
+    public const string ExpressionNestingTooDeep = "Calor0118";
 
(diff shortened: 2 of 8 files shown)


juanmicrosoft:
Review response — both blockers fixed in 2d1b6887, verified by reproducing your inputs as tests. Point-by-point:

**🔴 1 (uncatchable StackOverflow) — fixed at the recursion nexus, reproduced first.** Your diagnosis was exactly right, including where a naive fix would fail: the crash cycle is `ParseLispArgument → ParseParenExpressionOrInlineLambda → ParseLispExpression`, which never passes through `ParseExpression` — my first guard placement there did nothing and the repro still aborted the test host. The guard now anchors in `ParseLispExpression` (every paren-nesting cycle passes through it): a deterministic depth cap (128) emits one `Calor0118` diagnostic, swallows the sub-expression to bracket balance, and lets the outer frames unwind normally. `EnsureSufficientExecutionStack` backstops in `ParseExpression`/`ParseStatement`/`ParseLispExpression` convert any residual deep recursion (deep `§IF` chains, tighter-stack environments) into a *catchable* exception, and strict `Parse` converts that into a clean `Failed`. Tests: 50k-deep nesting through tolerant and strict parse (both fail gracefully with the process alive — previously exit 134), plus a 4000-level `§IF` nesting no-crash test. The cap survives the full 7,600-test suite, so no legitimate corpus code comes near it.

**🟠 2 (FIFO hang) — fixed via the stat gate.** Zero-length entries are never opened: FIFOs (and sockets) stat as size 0, and a genuinely empty regular file reads as `""` anyway, so the guard is behavior-preserving for real files. Applied to session load, refresh, and both `FileWriteTool` reads; enumeration additionally skips `Device` entries. A FIFO as the *write target* is treated as empty and the atomic rename replaces it without ever opening it. Tests: FIFO in the session tree (open completes, counted as a file), FIFO as write target (applies, becomes a regular file) — both with 30s watchdogs.

**🟡 Minors:** `PathGates` is now a fixed 64-stripe semaphore array — bounded forever, stripe collisions merely serialize. The pre-rename revalidation now also rejects a file *created* since the read (the clobber path you called out); `created` in the output keys off existence-at-read. M1 (stat-gate), external-writer rename race, and CWD-rooted no-session confinement remain documented stances, unchanged.

**Two corrections on the review's framing:**
1. *"798 re-bundles the entire session/file-write implementation"* — that was true when the review was written, but #797 has since merged to main (`b40c57a8`) and this branch is rebased on it; #798's diff against main is now only the D2.5 layer + these fixes.
2. *"the kickoff doc it cites isn't on main"* — `docs/plans/loop-m3-ws2.md` merged in #796 (`a5a54e4c`), and plan #747 merged as `762adb0b`. Both resolve on main today; if your tooling still reports otherwise it may be looking at a stale fetch.

Full suite after the fixes: 7,600+ tests, 0 failures, warnings-as-errors, Release rebuilt.

PR #799, opened 2026-07-25
feat: M3 — harness MCP registration, mcp-file gate opened, M-L2/M-L4 write stream (loop plan WS2/D4.2)

juanmicrosoft:
## Summary

M3 PR 4 (final slice, per `docs/plans/loop-m3-ws2.md` §3): the harness side of the WS2 mutation loop.

- **The #758 gate opens.** Its stated unblock condition — "WS2 D2.4 ships an MCP write tool and the harness registers the server" — is met by #797/#798. `run-pair.sh` now registers the calor MCP server for live `mcp-file` arms via `--mcp-config` + `--strict-mcp-config` (hermetic: the agent sees exactly one server, user-level MCP config cannot bleed into the arm). The server starts from `$ws/src`, so its no-session write-confinement root is the task tree itself.
- **Enforcement stays two-sided:** the PreToolUse hook blocking raw `Edit`/`Write` on `.calr` is unchanged, and the prompt constraint stands. Live `mcp-file` is calor-arm only and errors without `calor.dll` — a hook-blocked arm with no server would strand the agent and fabricate guaranteed-failure data. `--null-agent` plumbing mode is unchanged.
- **M-L2/M-L4 sourcing:** `calor_file_write` journals one `mcp-write/1` JSONL record per attempt (`CALOR_MCP_WRITE_LOG`) and archives rejected content (`CALOR_MCP_REJECT_DIR`) for the D4.6 reject-replay harness. Env comes from the server registration; absent → zero logging (telemetry can never fail the write path). `extract_metrics` aggregates the stream into `result.json` as `mcpWrites: {attempts, applied, rejected, healed}` — **M-L2(mcp-file) = applied/attempts**. The journal's `apply_verdict` stays reserved-null by design: one iteration can contain several write attempts and the shim cannot observe MCP traffic, so per-attempt granularity lives in the companion stream (documented in `loop-telemetry-schema.md`).

## Verification

- `bash -n` clean; full null-agent smoke run of the `mcp-file` arm green end-to-end (`mcpWrites: null` as expected — no server in plumbing mode).
- Live server driven over NDJSON against the Release binary: applied write (`safe`, file updated), rejected write (`breaking`, file untouched), both journaled, reject payload archived with the rejected content.
- `claude --strict-mcp-config` flag presence confirmed against the installed CLI.
- 3 new telemetry tests; all 40 MCP tests green; full suite green.

## What this unlocks

The WS2 exit criterion is now runnable: an agent completing a multi-edit E2E task exclusively through the MCP file tools, with M-L2 measured from the write stream and M-L4 reported from reject payloads (Annex A: reported-only below 20 rejects). That run spends API budget, so it's a deliberate next step, not part of this PR.

### DIFF
--- a/bench/phase0-agent-native/loop-telemetry-schema.md
+++ b/bench/phase0-agent-native/loop-telemetry-schema.md
@@ -55,6 +55,35 @@ censors at budget+1; censored fractions are reported per arm.
   `loop-baseline-ws1`) has no apply path, and the schema must be frozen
   **before** the treatment build exists so both arms emit identical shapes.
 
+## Companion stream: `mcp-writes.jsonl` (mcp-write/1)
+
+Live `mcp-file` arms register the calor MCP server (M3 PR 4), and
+`calor_file_write` journals **one record per write attempt** into
+`mcp-writes.jsonl` in the run's out dir (env `CALOR_MCP_WRITE_LOG`, set in the
+server registration). This is M-L2's per-attempt stream for the mcp-file
+mechanism — the journal's per-iteration `apply_verdict` remains reserved-null;
+per-attempt granularity lives here instead, because one iteration can contain
+several write attempts and the shim cannot observe MCP traffic.
+
+| Field | Type | Notes |
+|:------|:-----|:------|
+| `schema` | `"mcp-write/1"` | discriminator |
+| `ts` | string (ISO-8601 UTC) | attempt time |
+| `path` | string | canonical target path |
+| `verdict` | `"safe"` \| `"safe_with_warnings"` \| `"breaking"` | check-set verdict |
+| `applied` | bool | whether the atomic apply ran (`verdict != "breaking"`) |
+| `healApplied` | bool | auto-heal changed the content before checking (D2.5) |
+| `created` | bool | the write created the file |
+| `rejectPayload` | string \| null | path of the archived rejected-edit payload (`rejects/`, env `CALOR_MCP_REJECT_DIR`) — D4.6 replay input for M-L4 |
+
+Notes: revalidation races (file changed on disk mid-check → retry error) are
+not journaled — they carry no verdict information. `extract_metrics`
+aggregates the stream into `result.json` as
+`mcpWrites: {attempts, applied, rejected, healed}` (null when the stream is
+absent). **M-L2(mcp-file)** = `applied / attempts`. **M-L4** consumes the
+`rejectPayload` archives; below 20 rejects per epoch it stays reported-only
+(Annex A).
+
 ## Compatibility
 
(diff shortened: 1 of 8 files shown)


juanmicrosoft:
Review response — the main finding is addressed in 1efe2512, with the claude-spawned verification you asked for.

**🟠 1 (unpinned CWD assumption) — pinned and smoke-verified through claude's own spawning.** You were right that the confinement root rested on an unverified client behavior. `calor mcp` now takes `--root`, threaded to `calor_session_open` and `calor_file_write` as the confinement root; process CWD is only the documented fallback, and the option's help text tells harness/CI setups to pin it. `run-pair.sh` passes `--root $ws/src` in the server registration. And the integration path you called unexercised has now been exercised for real: claude launched from a directory *unrelated* to the task tree (so CWD propagation cannot be what makes it pass), `--mcp-config`/`--strict-mcp-config` with the exact config shape the harness generates, a live agent call to `calor_file_write` → applied inside the pinned root, file content updated on disk, `mcp-writes.jsonl` record written. One cheap pinned-Haiku call; the failure mode you predicted (silent fabricated-failure data from a wrong implicit root) is now structurally impossible rather than assumed away.

**🟡 Heal accounting — adopted into Annex A as additive revision A-1.1.** M-L2(mcp-file) is now explicitly defined as first-apply-**after-autoheal** validity; `extract_metrics` adds `appliedUnhealed` so the strict form is first-class in `result.json` (not an analyst exercise), and A-1.1 requires both forms in any cross-mechanism comparison. No thresholds changed; annex version bumped with a revision-log entry per its own discipline.

**🟡 Reject archive — bounded** at 200 payloads per server process; records keep flowing past the cap with `rejectPayload: null` (M-L4's adjudication floor is 20, far below).

**🟡 JSON escaping — fixed** as a side effect of item 1: the config is now built with `jq --arg`, so `$CALOR_CLI_DLL`/`$ws_out`/root paths are properly escaped.

**Two stale carried-forward items, for the record:**
- *"the FIFO-hang DoS from my #798 review"* — fixed in #798 itself (commit 2d1b6887, merged in bec9cf64): zero-length entries are never opened, with FIFO-in-session and FIFO-as-write-target tests. Your re-check may predate that commit.
- *"the kickoff/Annex authority still isn't on main"* — `loop-plan-v0.9.md` (762adb0b), `loop-m3-ws2.md` (a5a54e4c), and the Annex (2464a060) are all on main; this has been repeated across three reviews against a stale checkout, please re-fetch.

M1 (stat-gate) remains the documented BuildStateCache-semantics stance.

Full suite after the fixes: 7,600+ tests, 0 failures; `bash -n` clean; null-agent smoke unchanged-green.

PR #800, opened 2026-07-26
fix: harden run-pair watchdog — verify-and-escalate tree kill, unbuffered driver log

juanmicrosoft:
## Summary

The first WS2 exit-epoch attempt (`ws2-exit-e2e-001`, aborted) ran into an Anthropic API incident, and the incident exposed a real harness defect: the bash-fallback watchdog's one-shot `sleep && kill -9 -- -pgid 2>/dev/null` failed silently, letting run 1's agent run **105 minutes against a 900 s budget** — the exact spend-guarantee failure the #758 review item was supposed to close. pgid evidence from a later run shows claude does stay in the subshell's process group, so the mechanism was right but unverified and unlogged; whatever went wrong left no trace.

- **`kill_agent_tree`**: collect the descendant set *before* killing (a dead root reparents children), kill process group + collected tree, **verify** death, escalate to a pattern kill scoped to the run's unique mktemp workspace path (also reaps an orphaned MCP server). Every step logs to stderr — no more `2>/dev/null` hiding failures.
- **Deadline poll in the parent** replaces the separate watchdog process (nothing to lose or race).
- **`CALOR_P0_TIMEOUT_OVERRIDE`** test hook so the watchdog is exercisable in seconds.
- **`jq --unbuffered`** on run-epoch's result pipe: with the driver's stdout in a log file, block buffering held every per-run result line until epoch end while stderr streamed — the aborted epoch looked like it produced nothing while actually progressing.

## Verification

Synthetic test: stub `claude` that hangs with a child process (MCP-server shape), 15 s override → watchdog fired on all three invalid-run retry attempts, killed root **and** child (zero survivors verified), run counted invalid by the existing retry cap. `bash -n` clean on both scripts; null-agent path untouched.

The live WS2 exit epoch will be rerun from this branch once the API incident clears.


juanmicrosoft:
Review response — one evidence-based pushback, one acknowledgment, and a small hardening commit to follow.

**🟡 1 (escalation misses the real MCP server) — the factual premise is stale; the claim holds in the code as shipped.** The cited `run-pair.sh:302-303` (`args: [dll, "mcp", "--stdio"]`, workspace only in env) is the **pre-#799-review** revision. Your own #799 finding #1 changed it: the config now passes the confinement root in **argv** — `run-pair.sh:309`: `args: [$dll, "mcp", "--stdio", "--root", $root]` with `$root = $ws/src`. Direct evidence from the epoch running under this branch right now: the live server's command line is `…calor.dll mcp --stdio --root /private/var/folders/…/p0-W2-005-scheduler-calor-inB5az/src`. So `pgrep/pkill -f "$ws_path"` does match the production server (the `/private` symlink prefix on macOS is covered — the pattern matches as an unanchored regex inside the longer canonical path). Your underlying principle — pattern kills see argv, not env — is exactly right, and it's worth making the coupling explicit so a future config edit can't silently break the escalation: I'll add a comment at the `--root` config site and in `kill_agent_tree` documenting that the escalation depends on `--root` being argv-visible, plus a defensive sweep of `$ws_out`-scoped processes (the shim wrapper's argv carries that path). That commit lands after the currently-running epoch completes — the epoch executes `run-pair.sh` from this branch's working tree, and editing a script bash is mid-way through reading corrupts the running process.

**🟡 2 (hardening-without-diagnosis, live guardrail unverified) — acknowledged, and the live evidence is being generated as we speak.** The WS2 exit epoch rerun is executing under this branch's watchdog right now: runs 1–2 of W2-005 completed valid (taskSuccess, 0 escaped bugs, green in 1 iteration, all edits through MCP), and the API incident's tail has already exercised the invalid-run retry path live ("api error" attempt → retry → valid completion). If the watchdog ever fires live, it now leaves stderr evidence by construction. The epoch record will state whether the guardrail was exercised or merely armed.

**Trivia** — agreed on both; the ≤5 s overrun granularity is documented by the loop structure.

juanmicrosoft:
Follow-up commit landed as promised, plus the live evidence both reviews asked for — the epoch ran to completion under this branch's watchdog:

- **Live watchdog verification (your #2):** W3-004 run 3's attempts 1–2 hung during the API incident's tail; the watchdog fired at 600 s both times, killed the full trees (zero survivors verified post-epoch), logged to stderr, and the invalid-run retry cap counted the run as failure. The spend guarantee held under real conditions. Driver log committed with the epoch record.
- **Argv-coupling (your #1):** comments now pin the dependency at both ends (`--root` config site + `kill_agent_tree`), and the escalation additionally sweeps the run's out-dir path. Synthetic test re-run green after the change.
- **Epoch result:** WS2 exit criterion MET — 5/5 valid runs exclusively through MCP, M-L2(mcp-file) = 100 % (both A-1.1 forms), multi-edit = 4–5 declarations per transactional write on W3-004, M-L4 reported-only at 0 rejects. Full record in `epochs/ws2-exit-e2e-001/` + `VERDICT.md`.

This PR now carries: watchdog hardening + its live verification + the WS2 exit epoch record + the M3-complete status update. Merging it closes M3.

PR #801, opened 2026-07-27
docs: M4 kickoff — WS3 warm feedback scope, priced decisions, allowlist entry (loop plan M4)

juanmicrosoft:
## Summary

M4 kickoff record per loop-plan §3 kickoff discipline (same pattern as #796 for M3): `docs/plans/loop-m4-ws3.md`. Docs only — code audit showed M4 plans no new C# product source, so no `.calor-csharp-allowlist` entries.

**Priced decisions recorded:**
- **"Rebind + Roslyn workspace" descoped as written** — the emit path is C#-text-only (no Roslyn object exists in driver/watch/MCP), the Call 1 descope removed the consumers (D2.2/D2.3) warm bound state was for, and effect-summary reuse is already delivered where summaries are consumed (`BuildStateCache` 2.0 on the CLI/watch warm path; the session path has no consumer). D3.1 = warm *derived* state on the session parse cache: hash-verified cached-parse reuse on the write path's original side + a lazy per-file call-target index. Revisit trigger = the PP-L1 miss path itself.
- **Warm state keyed to the parse cache** — invalidated exactly with parse state (existing stat+hash gate); no second invalidation scheme, no watcher.
- **Check-set parity** — D3.1 is performance-only; warm paths must reproduce today's verdicts (parity-tested). Cross-module effect enforcement stays out (verdict change → v0.10).
- **PP-L1 adjudicates on the MCP write path**; `watch` rebuild is instrumented and reported, excluding the configured debounce window from the measured wall time.
- **`.calorproj` trigger from M3 fires and is declined** — fixture is a directory session; re-arms for v0.10.
- **Box re-confirmed at 2–3 wk** (down from 3 wk; rebind/Roslyn descoped).

**Slicing:** PR 2 = D3.1 warm derived state; PR 3 = D3.2 latency instrumentation (mcp-write/2 + machine schema, watch rebuild elapsed); PR 4 = D3.3 fixture + governance; PR 5 = M-L1 measurement + PP-L1 verdict + release-notes entry.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #802, opened 2026-07-27
feat: M4 — warm derived session state: cached-parse reuse + call-target index (loop plan WS3 D3.1)

juanmicrosoft:
## Summary

D3.1 per the M4 kickoff (#801): warm *derived* state on the session's parse cache, performance-only with verdict parity.

- **Write-path original parse reuses the session's cached parse** — `CheckAndApplyAsync` previously re-ran `ParseTolerant` over on-disk content the session already held parsed. `FileWriteTool.WarmOriginalParse` now serves the cached `ParseResult` when the session holds the file with an identical content hash (verified against the bytes just read, so a stat-preserving edit can never slip a stale AST in); cold fallback otherwise. Verdict-neutral: the cached result came from the same `ParseTolerant` over the same bytes.
- **Lazy per-file call-target index on `SessionFileState`** — computed at most once per parse state (`Lazy<T>`, thread-safe), invalidated exactly when parse state is invalidated (a reparse or applied write constructs a new state and with it a fresh lazy). `CheckProjectReferences` consumes the index instead of re-walking every neighbor's AST per call.
- **`ProjectSession.TryGetFile`** accessor for path-keyed state lookup.

Per the kickoff's priced decision, no session-side effect summaries (no consumer post-Call-1-descope) and no binder/Roslyn state.

## Tests

Three new tests in `SessionAndFileWriteToolTests`: hash-verified warm/cold parse selection, index computed-once + invalidated-on-reparse, and end-to-end verdict parity across repeated warm calls (breaking stays breaking; behind-the-back caller change flips the same write to applied). Full suite green (7,435 passed, 0 failed).

Existing C# files only — no allowlist exposure.

### DIFF
--- a/src/Calor.Compiler/Mcp/Sessions/ProjectSession.cs
+++ b/src/Calor.Compiler/Mcp/Sessions/ProjectSession.cs
@@ -160,6 +160,21 @@ internal sealed class ProjectSession
     }
 
     /// <summary>
+    /// The session's state for <paramref name="absolutePath"/>, or null when
+    /// the file is not part of the session. Lets the write path reuse the
+    /// cached parse instead of re-parsing on-disk content the session
+    /// already holds (WS3 D3.1).
+    /// </summary>
+    public SessionFileState? TryGetFile(string absolutePath)
+    {
+        var path = Path.GetFullPath(absolutePath);
+        lock (_sync)
+        {
+            return _files.TryGetValue(path, out var state) ? state : null;
+        }
+    }
+
+    /// <summary>
     /// Records the result of an applied write, so subsequent checks in this
     /// session see the new content without re-reading the file.
     /// </summary>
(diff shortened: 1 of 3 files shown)


juanmicrosoft:
Adversarial review round complete; the reviewer confirmed the core parity claim (warm reuse gated on exact path + SHA-256 vs same-call read; all downstream consumers read-only; Lazy/locking sound) and reported 3 low + 1 nit, all addressed:
1. (low) **Case-insensitive path miss**: `TryGetFile` now falls back to an OrdinalIgnoreCase scan on non-Linux platforms when the exact lookup misses — a client writing `MATH.calr` against an enumerated `math.calr` no longer silently defeats the warm cache. Safe by construction: callers hash-verify content before reuse, so a wrong-entry match can only fall back cold. The two adjacent pre-existing casing issues the reviewer noted are fixed under the same platform rule: `UpdateFile` replaces a case-variant entry instead of inserting a duplicate key, and `CheckProjectReferences`' self-skip compares case-insensitively on those platforms (kills the stale-self-entry false-"breaking").
2. (low) **`Refresh` 512 KB cap bypass**: the stat-changed branch now routes a file grown past the cap through the same oversize guard as `Load` — the "session refuses what a tool refuses" invariant holds for the file's whole session lifetime, not just at open.
3. (low) **Test gaps**: added a casing test that feeds `WarmOriginalParse` exactly the canonical path the integrated write path computes (so a normalization regression fails it), a healed-write warm-reuse test (`Assert.Same` on the stored parse after `healApplied`), and a grown-past-cap Refresh test.
4. (nit) Lazy exception caching: no action, as the reviewer suggested — the walker is deterministic; noted.
32/32 class tests green; solution builds clean.

juanmicrosoft:
Re-verification round: the reviewer confirmed findings 1/4/5 fully resolved (fallback safety argument sound; oversize guard correct; tests pin their claims) and flagged one **new low** introduced by my fix plus a documentation gap — both now addressed in the latest commit:
- **UpdateFile identity guard**: before reusing a case-variant key, `DirectoryHasExactEntry` checks whether the directory holds an entry under the exact written name — on a case-insensitive filesystem a differently-cased write lands on the existing entry (no exact-name entry appears → reuse variant key, no phantom duplicate), while on a case-sensitive volume a genuinely distinct file does have one (→ own key, **no clobbering of a different neighbor's state**, which my previous fix could do).
- **Per-OS-proxy caveat documented** at both platform-rule sites: `!IsLinux()` is a per-OS proxy for a per-volume property; the read path is hash-guarded against the mismatch, `UpdateFile` now carries the identity guard, and the self-skip's accepted residual (a distinct case-variant twin exempted from the best-effort veto on a case-sensitive volume) is stated in place.
- **Test added**: case-variant `UpdateFile` replaces instead of duplicating, with session state reflecting the write (canonical-path usage mirroring production).
33/33 class tests green.

PR #803, opened 2026-07-27
feat: M4 — edit→envelope latency instrumentation: mcp-write/2 + watch-rebuild/1 (loop plan WS3 D3.2)

juanmicrosoft:
## Summary

D3.2 per the M4 kickoff (#801), stacked on #802. Both loop surfaces now log edit→envelope wall time into the loop telemetry streams; measurement boundaries per the kickoff's priced decision.

**MCP write path (M-L1's adjudicating surface):**
- `Stopwatch` from tool-call receipt in `ExecuteCoreAsync` through the verdict record — covers session refresh, per-path gate wait, heal, parse, check set, and atomic apply.
- `mcp-write/1` → **`mcp-write/2`** (additive): `latencyMs` + phase breakdown `refreshMs`/`checkMs`/`applyMs` (`applyMs` bounded to the revalidate+rename block only; 0 on reject).
- `extract_metrics` is unaffected — it never filtered on the discriminator (verified).

**Watch path (reported, not adjudicating):**
- `Rebuild` is timed from start to NDJSON-envelope-written — the debounce window is excluded by definition (configured delay, not feedback cost) and recorded alongside as `debounceMs`.
- New **`watch-rebuild/1`** JSONL stream via env `CALOR_WATCH_REBUILD_LOG` (mirrors the `CALOR_MCP_WRITE_LOG` pattern, IO errors swallowed); `RebuildResult` gains `ElapsedMs`; status line now shows rebuild wall time. The shared envelope schema v1.1 is untouched.

**Bench (telemetry contract):**
- Machine schemas added: `mcp-write.schema.json` (accepts `/1` archives and `/2`; closes the no-machine-schema gap noted at kickoff) and `watch-rebuild.schema.json`. `rejectPayload` is optional in-schema because the MCP serializer drops null fields (verified against the ws2-exit-e2e-001 archives).
- `validate-telemetry.sh` now dispatches every record on its `schema` discriminator to the matching schema file (unknown → invalid); the hand-rolled fallback gains a generic flat-schema checker. Archived epoch journals re-validate green; a deliberately malformed record is rejected with precise errors.
- `loop-telemetry-schema.md` documents both streams and the compatibility story.

## Tests

New `WatchRebuildTelemetryTests` (elapsed reported, record written when configured, no telemetry otherwise — first direct `WatchSession` coverage); latency-field assertions added to the mcp-write logging test. Full suite green (7,438 passed, 0 failed). `run-pair.sh` untouched.

### DIFF
--- a/bench/phase0-agent-native/loop-telemetry-schema.md
+++ b/bench/phase0-agent-native/loop-telemetry-schema.md
@@ -55,7 +55,7 @@ censors at budget+1; censored fractions are reported per arm.
   `loop-baseline-ws1`) has no apply path, and the schema must be frozen
   **before** the treatment build exists so both arms emit identical shapes.
 
-## Companion stream: `mcp-writes.jsonl` (mcp-write/1)
+## Companion stream: `mcp-writes.jsonl` (mcp-write/2)
 
 Live `mcp-file` arms register the calor MCP server (M3 PR 4), and
 `calor_file_write` journals **one record per write attempt** into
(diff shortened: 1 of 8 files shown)


juanmicrosoft:
Adversarial review round complete; the reviewer verified the timing arithmetic, logging paths, /1 backward compat, and loop-telemetry/2 enforcement as sound, and reported 1 medium + 4 low + 2 nits — all addressed (plus the updated #802 base merged in):
1. (med) **Schema can now catch a writer that stops emitting the adjudicating measurement**: `mcp-write.schema.json` gained draft-2020-12 `if/then/else` — the four latency fields are **required** on `/2` and **forbidden** on `/1`. The hand-rolled fallback gained matching const-discriminated conditional support. Verified in BOTH modes against: a `/2` missing latency (rejected), a `/1` carrying latency (rejected), real `/1` archives (pass), real `/2` records (pass) — 3 valid / 2 invalid identically in jsonschema and stdlib modes.
2. (low) **Mode divergence on integer-valued floats fixed**: the hand-rolled checker now accepts `41.0` for `"integer"`, matching jsonschema draft-2020-12 semantics — same file, same verdict, either machine.
3. (low) **latencyMs boundary stated once, normatively**, in the telemetry doc: tool-handler receipt → verdict journaled; protocol parse, result-envelope serialization, and reject archiving are explicitly outside the clock (all µs-to-low-ms vs the 300 ms gate). The kickoff doc's "envelope serialized" wording will be aligned on #801.
4. (low) **watch-rebuild/1 doc edges fixed**: `rebuild` documented as a monotonic counter whose logged ordinals may skip (no-sources rebuilds consume an ordinal, write no record), and the `--format text` boundary (rebuild start → compile end, no envelope) is stated; M-L1 runs use `--format json`.
5. (low) **Deterministic latency-test assertions added**: phase-sum ≤ latencyMs on the applied path (pins mark ordering and point-of-sample — a swapped or pre-sampled implementation now fails), plus a reject-path `/2` record test (schema, applyMs ≈ 0, rejectPayload present, phase-sum invariant).
6. (nit) `WatchSession`'s rebuild-log lock is now process-wide (static), matching `FileWriteTool.WriteLogSync`.
7. (nit) **Stream purity restored** via `validate-telemetry.sh --expect <discriminator>` — a misrouted record type in a single-stream file is an error again.
35/35 affected tests green; solution builds clean.

juanmicrosoft:
Re-verification round: **7/7 findings confirmed RESOLVED, no new defects** — both validator modes now score the reviewer's 14-record malformed corpus identically, the /2-missing-latency and /1-with-latency regressions fail in both modes, and the boundary docs match the code's actual sample points. The one residual nit (reject-path `applyMs <= 1` could flake on a GC pause landing between the two clock samples) is fixed with a flake-safe 50 ms bound that keeps the intent. The updated #802 base (UpdateFile identity guard) is merged in; 36/36 affected tests green.

PR #804, opened 2026-07-27
feat: M4 — D3.3 latency fixture: 10k-line multi-module project, governance, timed-edit script (loop plan WS3)

juanmicrosoft:
## Summary

D3.3 per the M4 kickoff (#801). Everything under `bench/phase0-agent-native/latency/`; no product code. Independent of #802/#803.

**Committed fixture** (`fixture-10k/`): 106 files, **10,275 non-blank lines**, 1,366 functions across Main + 3 layers × 35 modules with strictly-downward cross-module calls (module-graph reference depth 3). Pinned: measurements run against the bytes in git, and the seeded generator (`generate-fixture.py`, seed 4537) exists for governed regeneration only.

**Governance in `fixture-10k/README.md`** (plan review m4's requirements): owner, regeneration policy (regenerate + re-baseline on minor version bumps, via reviewed PR), sample count (the full 230-edit script per measurement, ≥ the plan's 200), and content criteria vs corpus percentiles from `corpus-stats.py` (snapshot committed). One deliberate deviation recorded: corpus cross-module depth is **0** — no multi-module project exists in samples/ or bench pairs — while D3.3 mandates multi-module; depth 3 is the stated choice.

**Timed-edit script** (`edits/edit-script.jsonl`): 230 seeded edits — 200 safe constant bumps (chained; each `find` unique to its target function) + 30 breaking renames drawn only from flow functions with ≥1 caller, so every rename genuinely dangles. Application rule (replace-ALL, in order, from committed state) documented in the README.

**Acceptance evidence (run before commit):**
- all 106 files compile clean under `--enforce-effects` (cross-module enforcement active)
- `calor verify`: **4,098 obligations proven, 0 disproven, 0 unsupported**
- full edit-script simulation: all 230 find/replace invariants hold; the post-script state (200 safe edits applied) also compiles 100% clean

**Verifier finding, recorded in the README:** contract forms are restricted to the verifier's current proven surface (single-parameter inequality chains per the D1.5 `Outcomes/proven.calr` form). During generation I found that the verifier refutes (a) any `result`-referencing postcondition — `result` comes back unconstrained by the body; `samples/Verification/proven-contracts.calr` itself reports 7/10 "potentially violated", e.g. Identity (`§S (== result x)`, `§R x`) with counterexample x=0/result=-1 — and (b) parameter arithmetic that can overflow i32 bitvectors (that one is honest). (a) is pre-existing on main (reproduced with the Jul 25 Release binary) and deserves its own issue; the fixture README records the constraint and the re-baseline trigger if it gets fixed.


juanmicrosoft:
Adversarial review round complete; all 6 findings addressed (fixture `src/` and the edit script are **byte-unchanged** — regeneration verified; only scripts/README/snapshot changed):
1. (med) The calls/function README note is now **computed** from the actual density vs corpus percentiles — the generator can no longer emit a false band claim. Current honest value: 0.41, below corpus p50 (0.50), above p25; rationale stated.
2. (low) `corpus-stats.json` roots are now repo-relative (no machine paths/username in the committed snapshot).
3. (low) The corpus-scope choice (samples/ + bench pairs, excluding tests/TestData) is now a second explicitly recorded deviation in the README, with the n=43 coarseness consequence stated. (Also recorded in the kickoff doc on #801.)
4. (low) "sized near p90" reworded to the accurate p75–p90 band with the real rationale (file-count practicality), dropping the non-binding session-cap justification.
5. (nit) The dead-looking kind-assignment short-circuit is replaced by a single unconditional RNG draw per function — same stream, verified byte-identical output.
6. (nit) `FUNC_RE` now matches generic names (`Identity<T>`), dotted call targets contribute both components as depth-graph candidates (density counting unchanged — one entry per call site), and the Python-version determinism dependency is documented in the regeneration policy. The regex fix legitimately moved corpus calls/fn p50 from 0.57→0.50; snapshot and README regenerated together.

juanmicrosoft:
Re-verification round: **6/6 findings confirmed RESOLVED, no new defects.** Critical gate passed: fixture `src/` and the edit script are byte-identical across the fix commit (ml1-001 provenance intact), and byte-exact regeneration at the new tip independently **proves** the RNG-stream restructure is equivalent. The corpus p50 shift (0.57→0.50 from the generics regex fix) was independently reproduced and is consistently reflected in the snapshot, README table, and computed note. One residual nit acknowledged by scope (theoretical dotted-call depth-edge collision — no instance in the corpus).

Issue #807, opened 2026-07-27
Verifier refutes result-referencing postconditions with an unconstrained result — spurious counterexamples instead of unsupported


juanmicrosoft:
## Summary

Any postcondition that references `result` is reported **Disproven** with a counterexample in which `result` is not constrained by the function body at all. The flagship sample `samples/Verification/proven-contracts.calr` — whose name and design say every contract should prove — reports **7/10 obligations "potentially violated"** on main.

The starkest case is `Identity`:

```calor
§F{f005:Identity:pub} (i32:x) -> i32
  §S (== result x)
  §R x
```

reported as: `Postcondition may be violated in function 'Identity'. Counterexample: x=0, result=4294967295` — i.e. the solver was free to pick `result = -1` for a function whose body is `§R x`. Same shape for `Square` (`§R (* x x)`, `§S (>= result 0)`, "violated" at `x=0, result=-1`): body→result binding is missing from the translation, so the postcondition is checked against a free variable.

## Repro (main, reproduced with both Debug and the Jul-25 Release binary — pre-existing, not from the M4 changes)

```bash
dotnet build src/Calor.Compiler/Calor.Compiler.csproj
dotnet src/Calor.Compiler/bin/Debug/net10.0/calor.dll verify samples/Verification/proven-contracts.calr
# → Proven: 3, Disproven: 7 (all three postcondition-carrying shapes fail)
dotnet src/Calor.Compiler/bin/Debug/net10.0/calor.dll -i samples/Verification/proven-contracts.calr -o /tmp/pc.g.cs --verify
# → same 7× Calor0712, so both the verify command and compile-time --verify paths agree
```

Distinguish from the **honest** refutation class: postconditions over parameter *arithmetic* (e.g. `§S (>= (+ x y) 0)` under `§Q (>= x 0), §Q (>= y 0)`) are refuted via genuine i32 bitvector overflow (`x=y=2147483647`) — that is correct behavior and not this bug.

## What currently proves

[...]

Issue #755, opened 2026-07-23
Soundness: precondition satisfiability (∃) conflated with validity (∀) — runtime guard elided on 'Proven' preconditions


juanmicrosoft:
Found while reviewing #754 (pre-existing on main; the emitter has zero changed lines in that PR).

**The bug.** `Z3Verifier.VerifyPrecondition` checks *satisfiability* of the precondition (`SatIsProof` polarity: SAT ⇒ `Proven` — "some input satisfies it"), but `CSharpEmitter.Visit(RequiresNode)` treats `Proven` as "always holds" and **elides the runtime guard** (default `--contract-mode debug`):

```
§Q (> x 0)          →   public static int Half(int x) {
§R (/ x INT:2)              // PROVEN: Precondition always satisfiable: x > 0
                            return x / 2;   // ← no check; Half(-5) does NOT throw
                        }
```

"Satisfiable" (`∃x. x>0`) is not "valid" (`∀x. x>0`). A caller passing an argument that violates the precondition sails through with no exception — the defensive check the contract promised is silently deleted.

**Why it matters now.** #754 routes this path through the new `ProofOutcome` choke point (`ProofStatus.Proven` → `ContractVerificationStatus.Proven`) and positions that type as the single verification soundness boundary. The status assignment is faithful to the solver evidence; the unsoundness is the *emitter's interpretation* of precondition-`Proven` as elision-safe. Until fixed, precondition `Proven` must not be treated as trustworthy for guard elision.

**Suggested fix directions** (pick one):
1. Emitter: never elide precondition guards based on satisfiability — only postcondition `Proven` (a genuine ∀-proof via UNSAT-on-negation) justifies elision. Smallest, safest change.
2. Verifier: introduce a distinct status/flag for "satisfiable" vs "valid" so precondition results can't be conflated with implication proofs (e.g. only report `Proven` for preconditions when the precondition is a tautology).
3. Both: (1) immediately, (2) as the typed fix.

**Repro:** compile the snippet above with `--verify --contract-mode debug` and call `Half(-5)` — no `ContractException` is thrown; the generated C# contains the `// PROVEN` comment and no guard.

Surfaced by an external review of #754; filed separately since the defect predates that PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #808, opened 2026-07-27
docs: M4b kickoff — WS5 wedge probe scope + Annex A-1.2 PP-W1/M-W1 registration (loop plan M4b)

juanmicrosoft:
## Summary

M4b kickoff per loop-plan §3 discipline (pattern of #796/#801): `docs/plans/loop-m4b-ws5.md` + the gates-doc **Annex A-1.2** additive registration of M-W1 and PP-W1's frozen threshold — registered before any probe fixture exists or epoch runs, per D4.4.

**Probe design (priced decisions):**
- **N = 9 injected defects**, 3 per class: W5-A undeclared side effect (build-blocking `Calor0410`), W5-B violated scalar contract (**Debug runtime guard channel only** — the static-verify channel is excluded until #807 is fixed, since spurious unconstrained-`result` refutations make a verify-based catch indistinguishable from noise; postcondition-only per #755's `Proven`-precondition guard elision), W5-C effect laundered through a covered cross-module chain.
- **Excluded holes, listed and frozen with the threshold**: delegate-typed invocation (assumed-pure; pinned by `DelegateInvocation_*` tests) and override/dispatch effect-laundering (no effect-variance rule) — a defect requiring them disqualifies its fixture.
- **Catch = defect absent at declared-done** per a per-defect held-out probe test; both arms get identical specs and identical agent-visible smoke tests; the C# arm keeps its strategy-§0 full toolkit.
- **PP-W1 threshold (frozen at A-1.2): Calor−C# delta ≥ 3/9 defects**, per-defect majority across ≥3 runs/arm/pair; **zero-vs-zero = the pre-committed Call 2 kill signal**.
- **Feasibility by determinism replaces a paid D4.5 dry-run** (a stated priced decision): W5-A/C catches are deterministic build blocks; W5-B's bounded nondeterminism is absorbed by 3 runs + majority.
- **Spend**: 54 agent runs estimated well under the $1,500/epoch gates-doc ceiling; the epoch does not run until the user authorizes the figure via the spend-authorisation process.
- Non-overlap with machine-zone §7's red-team gate stated precisely (human spec-diff review vs compiler enforcement; different corpus and comparator).

**Slicing:** PR 2 = D5.1 probe pairs (`W5-*`) + per-test held-out outcome extraction + fixture acceptance checks; D5.2 = the probe epoch (user-authorized), with PP-W1 adjudication deferred to Call 2 at M5.

Docs only; no product code, no allowlist entries.

### DIFF
--- a/docs/plans/agent-native-gates.md
+++ b/docs/plans/agent-native-gates.md
@@ -99,8 +99,8 @@ After freezing, this document may be superseded only for a **documented empirica
 
 ## Annex A — Instrument metrics (loop plan v0.9, D4.4)
 
-**Annex version: A-1.1 (thresholds frozen at A-1.0, 2026-07-24; additive
-clarification A-1.1, 2026-07-25).** This annex is **additive-only
+**Annex version: A-1.2 (thresholds frozen at A-1.0, 2026-07-24; additive
+clarification A-1.1, 2026-07-25; additive PP-W1/M-W1 registration A-1.2, 2026-07-27).** This annex is **additive-only
 with respect to the main document**: no §1–§7 machine-adjudicable gate
 criterion references any metric defined here, and nothing here alters the
 C#-vs-Calor gate decisions those sections govern. The annex's own proof-point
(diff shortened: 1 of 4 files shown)


juanmicrosoft:
Authoring-time amendment: W5-C is re-priced as **intra-module** laundering. While building the cross-module fixture I found a pre-existing emitter gap — cross-module calls emit unqualified C#, so a multi-module reference solution can never link under MSBuild/csc, and the qualified spelling is Unknown:* under strict policy — filed as **#809**. Enforcement itself catches cross-module laundering fine (verified: Calor0410 'via cross-module call' fires under both CLI and MSBuild); the constraint is that no defect-free reference could ship. The intra-module chain exercises the same claim (caller's declaration violated, intermediate's honest, full chain diagnostic) on a path that emits working C#. Revisit trigger recorded: #809 fixed → cross-module variant may be added.

juanmicrosoft:
Adversarial review round complete; all findings addressed in the latest commit:
1. (HIGH) **Guard exemption shipped in this PR**: `scripts/check-calor-first-diff.sh` now structurally exempts `bench/` (like `tests/`) with rationale — probe pairs are definitionally half C#, and the ~700 pre-existing bench .cs only ever passed via the grandfather clause. This is the base-branch change #810 needs; #810's red calor-first-guard check goes green once this merges and #810 updates from main. The kickoff's false 'no allowlist entries' sentence is corrected to describe exactly this.
2. (MED-HIGH) **Registration timing honesty**: A.3's 'before any probe fixture exists' replaced with the truth — freezing preceded the epoch, authoring was concurrent, and W5-C was re-priced from an authoring-time discovery (#809) before the class definitions froze. The class definitions are now **inlined in the frozen A.2 row** instead of referenced from the mutable kickoff table; what remains results-blind (threshold, adjudication rule, scope guards, fallback — zero live runs observed) is stated as the protected property.
3. (MED) **D4.5 supersession recorded where it belongs**: a §10 revision-log entry in `loop-plan-v0.9.md` (not self-authorization in the kickoff), acknowledging the precedent that both prior dry-runs moved their measures. The determinism argument's honest scope (surfacing channel, NOT agent fix-behavior) is stated in both docs, and a **pre-registered decidability fallback** is added to the frozen row: per-defect outcomes not unanimous for ≥7/9 defects in either arm → PP-W1 reported-not-adjudicated (the PP-L4 pattern).
4. (MED) Annex blanket prose qualified: 'thresholds through A-1.1 frozen from dry-002; A-1.2 by the recorded argument'.
5. (LOW) PP-W1 row precision: **exactly 3** runs (odd by construction, no tie rule needed); **raw aggregated count, no §6.1 significance test** (9 binary clusters degenerate under bootstrap); zero-vs-zero defined post-majority; negative delta = clean miss, thesis-adverse, reported with kill-level prominence.
6. (LOW) **M-W1 added to A.1** with its true telemetry source (the `defect{}` object in result.json — explicitly not loop-telemetry/2).
7. (nits) Glob corrected to `W5?-*`; machine-zone §7 descriptor updated to the §12-amended form; the §3 spend-letter wrinkle acknowledged (wedge figure enters now, before the epoch, matching the M2 precedent).
All compiler-facing claims were verified by the reviewer and cleared (delegate hole, dispatch gap, #807/#755 reasoning, Calor0410 build-blocking, cross-module enforcement tests, fixture-set consistency).

Issue #809, opened 2026-07-27
Cross-module Calor calls emit unqualified C# — multi-module projects cannot link under MSBuild/csc


juanmicrosoft:
## Summary

A bare-name cross-module call (`§C{SaveSnapshot}` from module `Catalog` to a public function of module `Store`) passes the compiler front-end — cross-module effect enforcement resolves it correctly (`CrossModuleEffectEnforcementPass` charges the callee's effects and names the chain) — but the **emitted C# is unqualified** (`SaveSnapshot(...)`), and each module emits into its own namespace/class (`Store.StoreModule`, `Catalog.CatalogModule`). csc then fails with `CS0103: The name 'SaveSnapshot' does not exist in the current context`.

There is currently **no working spelling** for a cross-module call:
- bare name: front-end fine, emission broken (CS0103 at MSBuild/csc time);
- qualified `§C{StoreModule.SaveSnapshot}`: emission would be closer, but effect enforcement treats it as an unknown external → `error Calor0410: uses effect 'Unknown:*'` under the strict unknown-call policy (and it would bypass the cross-module registry's summary propagation).

## Repro

```bash
mkdir -p /tmp/xmod && cd /tmp/xmod
cat > Store.calr <<'CALR'
§M{m001:Store}
  §F{f001:SaveSnapshot:pub} (str:path, str:name) -> void
    §E{fs:w}
    §C{File.WriteAllText} §A path §A name §/C
CALR
cat > Catalog.calr <<'CALR'
§M{m002:Catalog}
  §F{f001:Ping:pub} (str:path) -> void
    §E{fs:w}
    §C{SaveSnapshot} §A path §A "x" §/C
CALR
# Front-end: fine (both compile, cross-module effects resolved)
dotnet <repo>/src/Calor.Compiler/bin/Debug/net10.0/calor.dll -i Store.calr -i Catalog.calr --enforce-effects
# Emitted Catalog.g.cs contains bare `SaveSnapshot(path, "x");` inside
# namespace Catalog / class CatalogModule → csc CS0103 in any MSBuild build.
```

Observed while authoring the WS5 probe pairs (M4b D5.1): a two-module fixture's defect was correctly blocked by `Calor0410` at build — which masked that its defect-free reference could never link. The W5-C probe class was restructured to intra-module laundering (equally covered by `EffectEnforcementPass` with the full chain diagnostic) as a result; the kickoff records this.

## Impact

[...]

PR #810, opened 2026-07-27
feat: M4b — D5.1 injected-defect probe pairs + harness probe support (loop plan WS5)

juanmicrosoft:
## Summary

PR 2 of the M4b kickoff (#808): the complete D5.1 injected-defect fixture set — **9 probe pairs, 3 per class** — plus the harness support the probe epoch needs. All fixture-facing decisions follow the kickoff's priced decisions and Annex A-1.2's scope guards.

**Pairs** (each: defective starter in BOTH arms with the same declared intent — Calor `§E`/`§S`, C# doc comments + NRT; arm-shared agent-visible smoke suite restricted to the starting surface; per-defect held-out `Probe_*` test; `defect.json` manifest; references defect-free + task-complete):
- **W5-A (undeclared side effect)**: W5A-001-report-audit, W5A-002-inventory-log, W5A-003-config-echo — hidden write inside a pure-declared function → Calor starter build-blocked by `Calor0410` with the call chain; C# starter builds green under NRT + `EnforceCodeStyleInBuild` and passes smoke.
- **W5-B (violated scalar contract)**: W5B-001-shipping-quote, W5B-002-loyalty-points, W5B-003-rate-limiter — boundary slip violating a declared `§S` on inputs the smoke suite exercises (smoke asserts only ranges true under both behaviors) → Calor starter compiles but smoke fails with `ContractViolationException`; C# starter passes the same smoke. Runtime-guard channel only, postcondition-only (per #807/#755, as registered).
- **W5-C (effect laundered through covered chain)**: W5C-001-catalog-cache, W5C-002-ledger-view, W5C-003-session-index (length-3 chain) — read-only-declared query reaches a write through intermediates whose own declarations are honest → `Calor0410` with the full chain; the violated declaration is the caller's. Intra-module per the kickoff amendment (#809 blocks cross-module references).

**Acceptance** (machine-checked, `check-probe-pair.sh`, committed): **63/63 checks green across all 9 pairs** — per-class starter asymmetry verified against the real toolchain (CLI + MSBuild `Calor.Tasks` path), C# starters never caught statically, all 18 references build green and pass held-out (including every probe test) and smoke.

**Harness** (`run-pair.sh`, validated with zero-spend null-agent runs):
- `materialize()` mounts the pair's `smoke/` as an agent-visible test project with its own starting-surface `SmokeShim` (same suite bytes both arms).
- `extract_metrics()` emits `defect: {id, class, probeTest, caught}` — `caught=true` requires the probe test to literally run and pass (`Passed: 1` in a filtered re-run; exit codes alone are unreliable: a zero-match filter or failed heldout build can exit 0). Non-compiling final state = not-caught, consistent with the all-failing held-out rule. `defect: null` on non-probe pairs and invalid slots.

Also: exemplar parity notes corrected to arithmetic that reconciles with the stated convention (both authoring reviews flagged the counts).

Merge order: after #808 (the kickoff + Annex A-1.2 registration this implements). Next: D5.2 probe epoch — runs from a committed branch only after user spend authorization (54 runs planned; per-epoch ceiling $1,500 per the gates doc).


juanmicrosoft:
Adversarial review round complete; both HIGHs and all six MEDIUMs addressed (latest commits on this branch + an A.1 semantics addendum on #808):
1. (HIGH) **Stale-DLL false catch killed**: the probe now runs only when the declared-done build succeeded (`final_build_ok` gate) — the reviewer's demonstrated scenario (defect fixed mid-run, last edit breaks the build → false `caught:true` off a stale Src.dll) now yields not-caught, consistent with the all-failing rule.
2. (HIGH) **Invalid-marker collision removed**: W5B-003 renamed `rate-limiter` → `request-quota` with all identifiers/prose scrubbed — an agent's natural summary can no longer substring-match the `"rate limit"` invalid-run marker and shred the slot's runs.
3. (MED) **C#-arm toolkit unforeclosed**: on probe pairs (defect.json present) the harness prompt now permits agent-written scratch tests in both arms (ungraded; sources in src/ are what's graded) — the strategy-§0 'agent-generated tests' channel is real again.
4. (MED) **Smoke integrity enforced, not asserted**: the PreToolUse hook now blocks Edit/Write under smoke/ in BOTH arms, and a materialize-time hash is re-checked at extract → `defect.smokeTampered`; A.1 (on #808) freezes tampered runs as invalid-for-M-W1, re-run per §0.2.
5. (MED) The W5-C deviation reference is now precise in all three defect.json manifests (kickoff §2 re-pricing + Annex A-1.2 inlined definition + #809) — the dangling 'kickoff amendment' pointer is gone (the amendment itself lives on #808, which this PR merges after).
6. (MED) **Parity convention normalized** across all 9 pairs (1 per function + 1 per conditional branch, module not counted): W5A-002 → 7/4, W5A-003 → 6/4, W5B-002 → 8/5, notes state the convention explicitly.
7. (MED) **Spec wording fixed in all 9**: 'name and signature unchanged, and its behavior as declared and documented' — the spec no longer orders preserving the defective behavior.
8. (MED) **Probe decoupled from task completion**: probe tests moved to `tests/probe/` in their own project compiled against the STARTING surface (SmokeShim) — fix-without-finishing now scores a catch, and the acceptance suite gained the presence-side check this enables (probe must FAIL against the defective C# starter). A.1 freezes the semantics incl. invalid-slot = not-caught.
Nits: scratch logs kept on FAIL; CLI compile runs on a temp copy (no in-place .g.cs, no interrupt stranding); `DOTNET_CLI_UI_LANGUAGE=en` pinned in both scripts; stale 'Store module' comment fixed.
Re-verified: **90/90 acceptance checks green across all 9 pairs** (up from 63 — the probe-vs-starter and probe-vs-reference checks are new), null-agent runs produce the new `defect{...caught, smokeTampered}` shape correctly. The reviewer's core-asymmetry verification stands: no default Roslyn/NRT/style signal catches any seeded defect; smoke suites exercise the trigger inputs while asserting only both-ways-true properties.

PR #811, opened 2026-07-28
M4b — D5.2 WS5 probe epoch ws5-probe-001: M-W1 measured (calor 9/9, C# 4/9, Δ +5)

juanmicrosoft:
## WS5 wedge-probe epoch (loop plan M4b / D5.2)

Live probe epoch measuring **M-W1** (injected-defect catch rate) per gates-doc **Annex A-1.2**. 9 W5 pairs × 2 arms × 3 runs = 54 runs, pinned `claude-opus-4-8` (program-consistency pin, decided with the user), raw edit mechanism, run from committed state.

### Result
| Arm | M-W1 | Unanimity | Run-level |
|-----|------|-----------|-----------|
| **calor** | **9/9** | 9/9 | 27/27 |
| **csharp** | **4/9** | 7/9 | 14/27 |

**PP-W1 input = Calor − C# delta = +5** (frozen threshold ≥ 3). Both arms ≥ 7/9 unanimous → the pre-registered **decidability fallback does NOT trigger** (measurement decidable). The wedge is concentrated in the **W5-B runtime-contract channel** (C# 0–1/3 vs calor 3/3 on all three), plus W5A-003 (effect defect) and W5C-003 (length-3 chain). Where both arms catch (W5A-001/002, W5C-001/002) the agent noticed the defect without the compiler forcing it → no delta.

**PP-W1 adjudication is DEFERRED to Call 2 (M5)** per the registration — this epoch produces the measurement only. Call 2 also needs the M5 comparison epoch.

### Catch channels (observational)
- **W5-A / W5-C → `Calor0410`** effect build-block (8/9 calor runs per class).
- **W5-B → `ContractViolationException`** from the emitted `§S` guard, thrown in the agent's first smoke run (starter compiles; C# carries the contract only as a doc comment, ships all 3).

### Harness fix included (commit 5f3fd839)
The `smokeTampered` integrity hash counted build-generated `smoke/obj/**/*.g.cs` (post-build re-check vs pre-build baseline), which would have flagged `smokeTampered:true` on every real run → all 54 invalid per frozen A.1. Fixed by excluding `obj/`/`bin/` at both find sites; genuine source-tamper detection preserved. Verified null-agent clean + previously-failing cell now `smokeTampered:false`.

### Run integrity
54/54 valid, **zero cap-exhaustions**; six failed attempts across four slots (three calor, one csharp) absorbed by the invalid-run retry cap — mostly transient API errors, one a watchdog SIGKILL; no defect majority affected. An earlier full attempt was aborted mid-way after a sustained Anthropic API incident capped 2 calor slots — discarded and re-run cleanly once the API recovered. Spend ≈ \$7.93 recorded (≈ \$15 incl. aborted attempts), ceiling \$1,500.

### Record contents
`pins.json` (+ 9-pair suite), `analysis.json` (machine-readable), `rollup.json`, `VERDICT.md`, and per-run journals/results/agent.json/envelope. `.gitignore` updated to exclude WS5 `probe/` + `.probe_final.txt` + `.smokehash` run-internals (analogs of the existing `heldout/` exclusions) so the record matches the ws2-exit shape.

Provenance: branch = main `267aee94` + smoke-hash fix `5f3fd839`; record commit `73bfc10a`.


juanmicrosoft:
## Adversarial review — completed, findings addressed (commit cfa01f09)

An adversarial reviewer recomputed every headline number from the 54 raw `result.json` files and re-checked the frozen A.1 rules. **Verified correct** (recomputed independently): M-W1 calor 9/9 & csharp 4/9 (majority per `invalid||smokeTampered ⇒ not-caught`), PP-W1 delta +5; unanimity calor 9/9 & csharp 7/9 (the two non-unanimous csharp defects are W5B-001 and W5C-003, both `[F,T,F]`); the decidability rule is inclusive ("≥ 7 of 9", agent-native-gates.md:183) so the fallback correctly does NOT trigger at csharp's exactly-7/9; run-level 27/27 vs 14/27; tokens (csharp 321/69,545, calor 812/247,304) and spend ($7.93) exact; pins `compilerCommit`/`modelPin` correct; 54/54 present, all `invalid:false`/`smokeTampered:false`; no `probe/`/`.smokehash`/`.probe_final.txt` leaked into the record.

**Two findings, both fixed:**

1. **MED — VERDICT run-integrity narrative was wrong.** "Four transient API blips (all calor arm)" contradicted the committed `invalid.txt` logs: there are **six** failed attempts across **four** slots — three calor (W5B-003/run-2, W5C-001/run-2, W5C-002/run-1) and **one csharp** (W5C-003/run-3) — and one was a watchdog SIGKILL (`rc=137`), not an API error. Corrected to match the data. No measured number changes (all 54 final runs valid; the affected csharp slot, W5C-003/run-3, is already the non-unanimous 1/3 miss).
2. **LOW — smoke-hash fix hardening note.** Added a comment documenting that excluding `obj/`/`bin/` from the integrity hash is not a tamper hole (the .NET SDK's `DefaultItemExcludes` drops `obj/`/`bin/` from the Compile glob, so a planted `.cs` there never enters the smoke assembly).

No aggregation errors, fabricated/missing runs, or record leaks were found. The M-W1 measurement stands: **calor 9/9, C# 4/9, Δ +5**; PP-W1 adjudication remains deferred to Call 2 (M5).

PR #805, opened 2026-07-27
feat: M4 — M-L1 measurement + PP-L1 verdict: PASS (loop plan WS3)

juanmicrosoft:
## Summary

Final PR of M4 (kickoff #801) — merging this closes the milestone. Now rebased onto merged main and containing only this PR's own content: the M-L1 measurement driver, the two run records, the CHANGELOG release-notes entry, and the kickoff record flipped to COMPLETE.

**Adjudicating record `ml1-002`** — run on **merged main** (#801/#802/#806/#804 all in, including every adversarial-review fix commit), per the re-run policy committed to in review:

| Surface | P50 | P90 | P99 | PP-L1 |
|---|---|---|---|---|
| MCP write (adjudicating) | **2 ms** | 3 ms | **8.84 ms** | **PASS** (≤300/≤1000) |
| watch rebuild (reported) | 27 ms | 30 ms | 33 ms | n/a |

All 230 verdicts matched the edit script (200 applied / 30 rejected, no heals); 431/431 telemetry records valid (plus `--expect mcp-write/2` stream purity); cold session open 112 ms, cold watch compile 382 ms; result.json carries commit + tree hash + dotnet SDK per the hardened driver. `ml1-001` is kept as the pre-merge record — the two agree within 1 ms on every percentile. Per plan §5, PP-L1's pass **permanently retires the direct-to-IL backend proposal**.

**Driver** went through two adversarial review rounds (see comment history): request deadlines with the documented one-response-per-request assumption, watch integrity assertions, heal-drift guard, provenance stamps, and a dirty-tree refusal scoped to tracked changes + latency-dir untracked files.

Merging this completes WS3's exit criterion (results published in the release notes whichever way they land — they landed at ~30–100× under threshold).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #812, opened 2026-07-28
M5 kickoff — comparison-epoch scope, arm A/B isolation, PP-L5/PP-L6 design (loop plan M5)

juanmicrosoft:
## M5 comparison-epoch kickoff (loop plan v0.9 milestone M5)

Pins the M5 design **before** any build or spend — matching the per-milestone kickoff discipline (M3/M4/M4b each landed a kickoff doc first). Nothing here changes a frozen threshold; all numbers are quoted from `agent-native-gates.md` (Annex A/D4.4) and `loop-plan-v0.9.md` (§5/§6/§10).

### What M5 measures
- **PP-L5** — does WS2+WS3 tooling reduce convergence cost? Frozen threshold: **≥ 15 % relative reduction in median per-pair tokens-to-green**, arm A vs arm B, simultaneous, ≥ 7 pairs × ≥ 5 runs/arm, one-sided cluster bootstrap α=0.05. Iterations-to-green is floor-bound (M2 dry-run: median 1, 94 % at floor) → observational only.
- **PP-L6** (release blocker) — harness-config invariance + neutral-task parity, arm A vs arm B.
- **PP-L3 is retired** (`loop-m2-baseline.md`) — no sub-epoch.

### Arm design (attribution, loop plan §10 C2)
- **Arm A** = `loop-baseline-ws1` (`4f235cdc`, WS1-only: envelope but no MCP write path, no warm feedback → raw edits, cold feedback).
- **Arm B** = arm A **+ WS2/WS3 merged in isolation** (MCP write path + warm feedback). **v0.9 HEAD is explicitly NOT arm B** — it carries unrelated M2/M4b/WS5 changes that would confound attribution.

### The two un-built prerequisites (flagged, not built here)
1. **Arm-B isolation build** — base `4f235cdc` + only the WS2/WS3 **compiler/runtime** delta (write path + parse mode #797/#798; warm state + instrumentation #802/#806). Docs/fixtures/harness/records stay at main per §4; all M2 measurement commits (#758/#794/loop-feasibility-dry-002) and M4b/WS5 excluded. Real git surgery; verify arm-B `src/` diff vs 4f235cdc and a clean Release build.
2. **Harness per-arm build-pin** — `run-pair.sh` loads `calor.dll` from one fixed path today; M5 needs each arm pinned to a different build (arm A baseline dll, arm B isolation dll), tasks/pins otherwise byte-identical.

### Guards frozen into the design
Neutral censored cap (40 %), differential censored cap (5 pp → reported-not-adjudicated), and the **underpowered-not-missed** rule (the 15 % MDE is a design-stage estimate over 7 clusters; materially-higher realized variance → reported underpowered, not a clean miss).

### Call 2 disposition
Adjudicated on **PP-L5 + PP-W1 together**. **PP-W1 already HIT at WS5 (delta +5, `ws5-probe-001`)** — not zero-vs-zero — so the pre-committed kill signal is not triggered; the program proceeds and PP-L5 shapes v0.10 (hit → continue; miss → freeze loop investment pending transcript analysis).

### Spend
**NOT authorized by this doc.** The live epoch is ≥ 70 full-loop agent runs (much heavier per run than the WS5 probe); it needs its own explicit authorization after the prerequisites are in place, under the \$1,500/epoch ceiling.

### Sequencing
0. this kickoff → PR (**here**) → 1. build arm B (report feasibility) → 2. harness build-pin + null-agent shakedown → 3. feasibility confirm → 4. spend-auth + live epoch → 5. adjudicate PP-L5/PP-L6 + Call 2.

### DIFF
--- a/docs/plans/loop-m5-comparison.md
+++ b/docs/plans/loop-m5-comparison.md
@@ -27,8 +27,11 @@ undetectable at any N), which is exactly why the primary measure moved to
 tokens-to-green.
 
 **§6.1 adjudication:** cluster bootstrap over pairs (runs nested within pairs),
-**one-sided**, α = 0.05, on the paired per-pair ratios against the 0.85
-threshold constant (= 15 % reduction). "Significant" always means this rule,
+**one-sided**, α = 0.05, on the paired per-pair ratios against the threshold
+constant (gates §6.1). The gates doc states the threshold only as "≥ 15 %
+reduction"; **operationalized**, that is a paired arm-B/arm-A ratio ≤ **0.85**
+(= 1 − 0.15) — the numeric constant and the B/A orientation are our
+operationalization here, not frozen text. "Significant" always means this rule,
 never eyeballing.
 
 ## 2. Arm design (attribution — loop plan §10 C2, the hard rule)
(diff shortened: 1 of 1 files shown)


juanmicrosoft:
## Adversarial review — completed, findings addressed (commit 4aa94c4d)

A reviewer checked every factual claim against the frozen sources. **Verified correct** (no change): the PP-L5 threshold (gates Annex A:182, word-for-word), the M-L5 definition + paired-ratio rule, both censored caps (40% neutral → invalid; 5pp differential → reported-not-adjudicated, correctly distinct and scoped), arm-A identity (`loop-baseline-ws1`=`4f235cdc`=#757 WS1 exit), PP-L3 retired (`loop-m2-baseline.md`), the Call 2 criterion (loop plan §6.2), PP-W1 = +5 HIT (ws5-probe-001), pairs W2×5+W3×4+N1×4, the $1,500 ceiling, the underpowered-not-missed fallback, and the arm-A-no-MCP-server consistency (harness only hard-errors on missing dll when mechanism=mcp-file; arm A runs `raw`). `#803` correctly absent from the range.

**Two MED + one LOW, all fixed:**

1. **MED — invented "0.85" constant.** The doc presented `0.85` as if quoted, but it appears in no frozen source (the gates doc states only "≥ 15 % reduction"). Reworded to flag `0.85` and the arm-B/arm-A ratio orientation as *our operationalization*, not frozen text — restoring the doc's quote-only promise.
2. **MED — arm-B recipe imprecise / contradicted §4.** It undercounted M2 (three commits — #758, #794, `loop-feasibility-dry-002` — not one) and swept non-compiler commits (docs #796/#801, harness #799/#800, fixture #804, verdict record #805) into the build delta, which §4 says stay at main. Reframed: arm B's `calor.dll` needs only the WS2/WS3 **compiler/runtime** delta (#797/#798 write path + parse mode; #802/#806 warm state + instrumentation), to be verified by diffing arm-B `src/` vs `4f235cdc`; all M2 records excluded.
3. **LOW — "single fixed path"** → "current checkout's build (Release-preferred, Debug-fallback) via `CALOR_CLI_DLL`, no per-arm build-pin."

No HIGH findings; no science-corrupting error. The design (attribution, guards, Call 2) stands; these were precision/fidelity tightenings before the arm-B recipe is trusted in the follow-up build step.

PR #813, opened 2026-07-28
M5 harness — --calor-dll per-arm build-pin for the A/B comparison epoch (loop plan M5)

juanmicrosoft:
## M5 harness prerequisite: per-arm build-pin

`run-pair.sh` loaded `calor.dll` from the current checkout only (Release-preferred, Debug-fallback via `CALOR_CLI_DLL`). M5's PP-L5 needs **two compiler builds run against identical tasks/pins/fixtures** — arm A = `loop-baseline-ws1` (raw edits, cold feedback), arm B = baseline + WS2/WS3 isolation (mcp-file, warm feedback).

**Change:** add `--calor-dll <path>` overriding `CALOR_CLI_DLL` (which drives envelope generation, `ids index`, and — for `mcp-file` arms — the calor MCP write server). Unset keeps the existing auto-detect. A missing path errors explicitly. One localized edit (default + arg-parse case + detection override + usage string); no behavior change when the flag is absent.

### Shakedown (zero live-agent spend, null-agent) — all pass
Built both arms and ran the null-agent path pinned to each:
- **Discriminator** — baseline (arm A) `calor mcp --help` has **0** `--root`; isolation (arm B) has **1**: the two builds are genuinely different (arm A lacks the WS2 write path).
- **Bad path** — `--calor-dll /nonexistent` → exit 2, `ERROR: --calor-dll path not found`.
- **Arm A** (baseline dll, `raw`) → `taskSuccess:true, envelopeValidAll:true`.
- **Arm B** (isolation dll, `mcp-file`) → `taskSuccess:true, envelopeValidAll:true`.
- **Routing proof** — arm-A run internals (`.shim/dotnet`) reference the baseline worktree dll; arm-B's reference the main/isolation dll; **zero cross-leakage** (arm A never touches the main path, arm B never touches the worktree path).

### Context
Arm-B isolation build verified earlier (branch `m5-arm-b-isolation` @ `122851ae`; `src/` diff vs baseline = exactly the 12 WS2/WS3 files; note arm B's compiler is byte-identical to main's since M2/M4b/WS5 never touched `src/`). Arm A builds from the `loop-baseline-ws1` worktree (Z3 native seeded from main — identical across arms — so it compiles in seconds).

Follows the M5 kickoff (#812). Next after this lands: pre-run feasibility confirm → **spend authorization** → the live ≥70-run epoch. Spend is **not** requested here.

### Known follow-up (out of scope here)
`run-epoch.sh` iterates `calor`/`csharp` and does not forward `--calor-dll` or a per-arm mechanism, so it cannot yet express the M5 A/B (two calor builds: raw/baseline vs mcp-file/isolation). Driving M5 needs either two `run-pair.sh` invocations per pair or a small M5 orchestrator — the next step after this run-pair change lands. Also: `--calor-dll` pins the CLI/MCP-server/envelope build, not the arm's MSBuild compile (Calor.Tasks/Sdk/emitter, bound to the checkout); that is exactly correct for M5 because WS2/WS3 touched none of those (empty diff `loop-baseline-ws1..main`), but a codegen-changing workstream would need per-arm checkouts.

### DIFF
--- a/bench/phase0-agent-native/run-pair.sh
+++ b/bench/phase0-agent-native/run-pair.sh
@@ -90,6 +90,7 @@ ITERATION_BUDGET=10
 TIMEOUT_SECS=600
 EXEMPLAR_FILE=""
 EDIT_MECHANISM="raw"
+CALOR_DLL_OVERRIDE=""
 
 while [[ $# -gt 0 ]]; do
     case "$1" in
@@ -100,11 +101,12 @@ while [[ $# -gt 0 ]]; do
         --null-agent) NULL_AGENT=1; shift ;;
         --exemplar) EXEMPLAR_FILE="$2"; shift 2 ;;
         --edit-mechanism) EDIT_MECHANISM="$2"; shift 2 ;;
+        --calor-dll) CALOR_DLL_OVERRIDE="$2"; shift 2 ;;
         *) echo "Unknown arg: $1" >&2; exit 2 ;;
     esac
 done
 
-[[ -n "$PAIR_DIR" && -n "$ARM" ]] || { echo "Usage: --pair <dir> --arm calor|csharp [--runs N] [--null-agent] [--exemplar <file>] [--edit-mechanism raw|mcp-file|mcp-node] [--out <dir>]" >&2; exit 2; }
+[[ -n "$PAIR_DIR" && -n "$ARM" ]] || { echo "Usage: --pair <dir> --arm calor|csharp [--runs N] [--null-agent] [--exemplar <file>] [--edit-mechanism raw|mcp-file|mcp-node] [--calor-dll <path>] [--out <dir>]" >&2; exit 2; }
 [[ "$ARM" == "calor" || "$ARM" == "csharp" ]] || { echo "--arm must be calor|csharp" >&2; exit 2; }
 case "$EDIT_MECHANISM" in
     raw) ;;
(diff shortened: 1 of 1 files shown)


juanmicrosoft:
## Adversarial review — completed, findings addressed (commit 46e32a61)

The reviewer verified the flag is mechanically clean: **override completeness** (every consumer reads the single `$CALOR_CLI_DLL` — mcp-config, `ids index`, envelope gen, both gates), **ordering** (override runs after arg-parse, before the gates), **regression when unset** (else-branch byte-identical to the old loop), **mcp-file live server uses the pin**, **path edge cases** (relative/spaces/symlink all safe), **`set -u` safety**, and the **bad-path exit 2**. No wrong-build leak on the edit path.

**Four findings; all addressed:**

1. **HIGH→precision — pin scope.** `--calor-dll` pins the CLI/MCP-server/envelope build, **not** the arm's `dotnet build` (which binds `Calor.Tasks`/`Sdk`/`Runtime`+emitter to the checkout). The reviewer flagged this could "measure a build against itself." **Verified it's exactly correct for M5:** the diff of `Calor.Tasks`/`Calor.Sdk`/`Calor.Runtime`/`CodeGen` between `loop-baseline-ws1` and `main` is **empty** — WS2/WS3's entire `src/` delta is in the MCP/session/envelope/parse-tolerant surface the pin governs. So both arms compile identically (codegen isn't the variable) and the whole arm-A-vs-arm-B delta is captured by the pin. Comment reworded to say this precisely, with the caveat that a codegen-changing workstream would need per-arm checkouts.
2. **MED — provenance.** `result.json` now stamps the resolved `calorDll` + `editMechanism`, so an M5 record can prove arm A ran the baseline build and arm B the isolation build (`pins.json` records only a single git HEAD).
3. **MED (scope) — no epoch-level caller.** `run-epoch.sh` can't yet drive the two-build A/B; flagged as the next follow-up in the PR body (out of scope for this run-pair-only change).
4. **MED/LOW — bogus-pin guard.** A path that exists but isn't a runnable calor CLI now errors (`dotnet <dll> --help`), preventing a live mcp-file arm from registering a dead MCP server and stranding the agent — the exact hazard the mcp-file gate guards. Re-verified: a text file named `.dll` → exit 2.

No wrong-build leak on the M5 measurement path. The flag is ready; the epoch orchestrator (Finding 3) is the next step.

PR #814, opened 2026-07-28
M5 — comparison-epoch orchestrator + analyzer (two-build A/B driver, loop plan M5)

juanmicrosoft:
## M5 comparison-epoch orchestrator + analyzer

Closes the #813 follow-up: `run-epoch.sh` iterates calor/csharp and can't express the M5 two-build A/B. Two new files under `bench/phase0-agent-native/`.

### `run-m5-epoch.sh` — the driver
One simultaneous A/B epoch over two calor **builds** against identical pairs/pins, via the `--calor-dll` build-pin (#813):
- **arm A** = `loop-baseline-ws1` build → `--edit-mechanism raw` (baseline, cold)
- **arm B** = baseline + WS2/WS3 isolation → `--edit-mechanism mcp-file` (warm)

Both `--arm calor`; the build + mechanism delta **is** the WS2/WS3 tooling PP-L5 measures. Only the pinned build differs — pairs/specs/held-out/harness are the current checkout, so tasks/pins are byte-identical across arms.

**Pre-flight guards** against the epoch-voiding case where both arms silently share a build: distinct dll paths **and** the WS2 `mcp --root` discriminator (arm A must lack it, arm B must have it) — fail-fast on a swap or wrong build. `pins.json` records **both** build commits + dll paths + mechanisms + the frozen PP-L5 (0.85) / PP-L6 config and the warm/neutral pair sets.

### `m5-analyze.py` — the adjudicator
- **PP-L5**: median paired **tokens-to-green** ratio (armB/armA) on the warm set (W2×5+W3×4), **one-sided cluster bootstrap** vs 0.85 (gates §6.1), with the 5pp differential-censored guard → reported-not-adjudicated when <2 warm pairs or censored-diff >5pp.
- **PP-L6**: neutral (N1) iterations-to-green parity + the 40% censored cap (config invariance is enforced per-run by `check_pins`).

### Null-agent shakedown (zero spend) — passed
Ran 1 warm + 1 neutral pair × both arms: preflight passed, both arms routed to their pinned builds (arm A → baseline worktree dll, arm B → main dll, **verified distinct**), `pins.json` carried both commits (arm A `4f235cdc`, arm B `f51fe2c4`), and the analyzer ran cleanly on the collected results (correctly reporting "not decidable" on the degenerate null data rather than crashing).

### Usage / next steps
```
CLAUDE_MODEL=claude-opus-4-8 ./run-m5-epoch.sh --epoch m5-compare-001 \
  --arm-a-dll <baseline-build>/calor.dll --arm-b-dll <isolation-build>/calor.dll [--runs 5]
```
After this lands, the remaining gates before the live epoch are: pre-run feasibility confirm and **spend authorization**. Spend is **not** requested here.

### DIFF
--- a/bench/phase0-agent-native/m5-analyze.py
+++ b/bench/phase0-agent-native/m5-analyze.py
@@ -3,128 +3,166 @@
 
 Reads an m5-comparison epoch dir (produced by run-m5-epoch.sh) and computes:
   - PP-L5: median paired tokens-to-green ratio (armB/armA) on the warm set,
-           one-sided cluster bootstrap vs the 0.85 threshold (gates Annex A / §6.1).
-  - PP-L6(b): neutral-set iterations-to-green parity (armB/armA), same bootstrap.
-  - Censored caps: 40% absolute (neutral) and 5pp differential (loop plan §4/m6).
-Writes analysis.json and prints a report. PP adjudication semantics are frozen
-in agent-native-gates.md — this only computes them.
+           adjudicated by the gates §6.1 decision rule.
+  - PP-L6(b): neutral-set iterations-to-green parity (no significant regression),
+              same bootstrap, plus the censored caps.
 
-M-L5 tokens-to-green = per-run agent output tokens; per-pair means, then the
-median of paired per-pair ratios (gates §2). A run is censored if invalid or
-never-green (iterationsToGreen > budget). Bootstrap is seeded for reproducibility.
+Frozen definitions (agent-native-gates.md), implemented here — not changed:
+  - M-L5 tokens-to-green = per-run agent output tokens; per-pair means, then the
+    median of paired per-pair ratios (§2).
+  - §6.1 decision rule: cluster bootstrap over pairs, RUNS NESTED WITHIN PAIRS,
+    one-sided, alpha=0.05. Pass iff the point estimate meets the threshold AND
+    the one-sided 95% CI excludes zero effect (ratio = 1.0). No third outcome.
+  - Censored: 40% absolute cap on neutral tasks (either arm) invalidates; a 5pp
+    differential (loop plan §4/m6) makes PP-L5 reported-not-adjudicated.
+
+A run is censored if invalid or never-green (iterationsToGreen > budget). The
+two-level bootstrap is seeded for reproducibility.
 """
 import json, sys, glob, os, random, statistics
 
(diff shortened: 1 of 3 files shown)


juanmicrosoft:
## Adversarial review — completed, findings addressed (commit 390a9ed7)

The reviewer verified metric fidelity, censored plumbing, arm-label mapping, pins wiring (both commits, annotated-tag `^{commit}`), CLAUDE_MODEL propagation, pair-set globbing (N1-004 RETIRED correctly excluded), and the valid-run jq hygiene — all correct. Four findings; all fixed:

1. **HIGH — a surviving-invalid run aborted the whole epoch.** `write_invalid_result` omitted `calorDll`, so the driver's `.calorDll | split("/")` hit a `null` → jq exit 5 → `pipefail`+`set -e` killed the epoch at the *first* invalid run, non-randomly truncating the dataset (and it was **latent through the null-agent dry run**, which is always valid). Fixed: stamp `calorDll`+`editMechanism` into `write_invalid_result`, and defensively guard the driver jq (`.calorDll // ""`, `.tokens.output // 0`). This was the finding most likely to corrupt a live PP-L5.
2. **MED — decision rule now faithful to gates §6.1.** The rule "stated in full" is: pass iff the point estimate meets the threshold **and** the one-sided 95% CI excludes **zero effect** (ratio 1.0). My old rule required the whole CI < 0.85, erring toward a **false miss**. Now: `hit iff point ≤ 0.85 AND P(bootstrap median ≥ 1.0) < 0.05`.
3. **MED — bootstrap is now two-level.** Cluster over pairs **with runs resampled within each resampled pair** (§6.1 "runs nested within pairs"), propagating within-pair variance instead of collapsing each pair to a point ratio (which understated variance, biasing toward a false hit).
4. **LOW — PP-L6 is now adjudicated** (regression-significance test + 40% censored cap + 5pp differential → pass/fail), not merely reported.

**Validated on synthetic epochs:** real+significant reduction → `HIT=True`; no effect → `HIT=False`; >5pp censored differential → reported-not-adjudicated; an injected invalid run is excluded from the means, counted censored, and does not crash; empty sets degrade to `None`. (Also fixed an empty-set crash the rewrite introduced.)

The one deferred item is the LOW arm-B `--root` discriminator brittleness — it **fails safe** (aborts rather than mis-running), so left as-is. This PR now also touches `run-pair.sh` (merged in #813) for the `write_invalid_result` provenance fix.

PR #827, opened 2026-07-30
G5 epoch driver — guarantees-probe-001 orchestrator + adjudicator

juanmicrosoft:
run-guarantees-epoch.sh: one A/B epoch over the frozen A-1.2 fixture set,
both arms calor+raw, per-arm-NATIVE products:
  arm A calor-v09-control  = tag guarantees-baseline-v0.9 PRODUCT CHECKOUT
    (template/Tasks/Sdk/Runtime/emitter/CLI all from the worktree — #826
    changed Calor.Tasks/Sdk.targets/template, so --calor-dll-only pinning
    is insufficient per run-pair.sh's own rule), verify gate ABSENT
  arm B calor-v10-verify   = current checkout, gate armed via
    CALOR_P0_VERIFY_EXPECTED=1 + CALOR_P0_VERIFY_GATE=1
Preflight: worktree HEAD must equal the tag; --weakening-check presence is
the build discriminator (v0.9 lacks it); driver-level refuted-canary on the
treatment CLI; inherited gate env is unset so the control arm cannot drift.
pins.json records both commits, roots, dlls, and per-arm gate config.

run-pair.sh gains two flags:
  --arm-repo-root <path>  binds the workspace PRODUCT (template
    instantiation, Tasks-libz3 pin) to a per-arm checkout while the
    measurement harness runs from the current checkout; the template file
    itself comes from the arm's checkout (per-arm-native includes the
    template, which is where the CalorVerify wiring lives)
  --arm-label <label>     names arms explicitly — this epoch runs the SAME
    arm/mechanism twice, so the default labels would collide

guarantees-analyze.py: the frozen instruments —
  M-G3 earliness (build-proof > build-block > runtime-guard >
    caught-unattributed > missed; majority of exactly 5; invalid = missed;
    build-proof = journal Calor0711/0712 on the seeded declarationId;
    runtime-guard from agent.json only, never .probe_final.txt)
  M-W1 continuity copy (exactly-5 majority)
  M-G4 weakening per valid W5-B run, both arms, via the v0.10 CLI's
    --weakening-check on the archived final-src (A-1.3.1 two-leg;
    indeterminates stay in the denominator; >20% => reported-not-adjudicated)
  PP-G3 legs a+b (leg-b joint predicate: build-proof AND probe-pass AND
    intactOrStrengthened DETERMINATE, >=3 of 5 slots, >=2 of 3 defects)
  PP-G4 legs a+b (itg two-level cluster bootstrap over all 9 pairs,
    regression = ratio>1, fail iff lower95 > 1.0, trio medians reported;
    weakening excess <= 3 runs absolute, 12-valid floor)
Seeded declaration extracted mechanically from the starter fixture (the
§S-carrying function).

Null-agent shakedown (1 pair, both arms): control surfaced as
caught-unattributed, treatment as build-proof (journal Calor0712 with
declarationId), M-G4 subprocess verdicts determinate, PP-G4 floors
correctly report not-adjudicated at shakedown scale.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


juanmicrosoft:
Adversarial review closed MERGE-CLEAN after one fix round (C1/C2/M1/M2/M3 + minors, all FIXED-VERIFIED empirically; annex amendment A-1.3.2 checked clause-by-clause against the implementation).

Disclosed residual (L, non-blocking): if the archived declared-done source contains the seeded declaration in TWO files (only reachable through a non-compiling final state — duplicate declarations are CS0101), resolve_final prefers the original basename, then first-containing alphabetically; no rule can know which copy the build consumed. Recorded here per the review's recommendation.

Issue #785, opened 2026-07-24
P0: Make effect enforcement symbol-resolved, exhaustive, and fail-closed


juanmicrosoft:
## Summary

Make effect enforcement symbol-resolved, exhaustive, and fail-closed under strict mode, and generate every effect surface from one authoritative taxonomy.

## Evidence and confirmed reproduction

- A bare-name pure-method allowlist includes mutators such as `Add`, `Remove`, `Clear`, `Sort`, and `StringBuilder.Append`.
- Unresolved receiver calls inherit empty effects by method name.
- Unknown constructors start with no allocation/unknown effect; constructor bodies are not fully checked or linked.
- Delegate/lambda/object-initializer and unhandled-node effects can disappear.
- Dotted calls can be reduced to bare local names.
- Confirmed: a pure function calling `items.Add(1)` passes strict enforcement.
- Effect registry, parser expansion, `EffectSet` surface parsing, manifests, and docs recognize different vocabularies; unknown source codes can be silently mapped.
- CLI/watch/SDK defaults and cross-module behavior are inconsistent.

## Required implementation

1. Resolve calls by symbol/signature, including receiver type, overload, constructor, delegate, and generic instantiation.
2. Charge allocation, mutation, disposal, and unknown effects explicitly.
3. Infer/check constructor, lambda, delegate, event, object-initializer, and collection-mutator effects.
4. Remove bare-name purity conclusions unless attached to a resolved immutable API signature.
5. Make unhandled/unresolved calls `Unknown`; strict mode must fail closed.
6. Generate parser codes, `EffectSet`, manifests, formatters, docs, and diagnostics from one taxonomy table.
7. Reject unknown/typo effect codes in source.
8. Define one consistent default policy across CLI, watch, SDK, MSBuild, MCP, and cross-module analysis.
9. Add exhaustive node and API-signature conformance tests.

## Required regression coverage

- Mutable versus immutable APIs with the same method name.
- Collection, StringBuilder, constructor, object initializer, delegate, lambda, event, and disposal effects.
- Overloads and dotted receiver resolution.
- Unknown external calls in strict/permissive modes.
- Every effect taxonomy code across source, manifest, parse, emit, and docs.
- Typo/unknown source effects.
- Cross-surface default-policy equivalence.

## Success metrics / definition of done

- The `items.Add(1)` reproduction fails in a pure function under strict mode.
- Zero production effect decisions rely on a bare method name without resolved API identity.
[...]

PR #842, opened 2026-08-02
feat: WS-W2 strictness batch — close the five effect-soundness holes

juanmicrosoft:
Implements WS-W2 from the v0.11 wedge plan (docs/plans/wedge-plan-v0.11.md §2). Merge-order gate satisfied: W1 Slices 1+2 on main; PP-W5 registration frozen results-blind in #839 BEFORE this merge (this PR is part of the epoch treatment arm).

## Deliverables

- **D-W2.1 — delegate invocation = error (Calor0418)**: the "assume pure" no-dot branch is gone. Bare-name targets resolving to a value (param, §B binding, field) invoked ⇒ error (warning only under the explicit `--permissive-effects` waiver); free names route through the unknown-call chain (0411 + worst-case ⇒ 0410) — fail-closed, never silently pure. Cross-module bare calls preserved via the driver's unambiguous pre-parse map plumbed into the pass.
- **D-W2.2 — effect variance, both legs**: override §E ⊆ base §E (**Calor0420**), implementation §E ⊆ interface §E (**Calor0421**), absent §E = pure. Call-site leg (in-module subset): calls through receivers statically typed as in-module interfaces/classes charge the declared §E — dispatch-sound precisely because 0420/0421 pin implementations to subsets. Residual (chained/external/property receivers) falls through to the existing fail-loud chain.
- **D-W2.3 — interop → Assumed (Calor0419), genuinely propagating**: assumption provenance (§RAW/§CS, conversion fallbacks, expression-target calls, unknown constructs, external-base overrides) with transitive closure over the reverse call graph; warning by default, **error under --strict-effects**, still disclosed under permissive.
- **D-W2.4 — mutator purge (#785)**: Add/Remove/Clear/Insert/Sort/Append/CopyTo/Log etc. purged from KnownPureMethodNames; untyped receiver now fails loud (0411 + 0410), typed receivers resolve via manifests with new generic-name normalization (`List<i32>` → ``List`1``). #785 repro dead both ways, pinned.
- **D-W2.5 — CLI default flip**: `--enforce-effects` defaults true on build + watch, `--no-enforce-effects` opt-out; ends the CLI/SDK split-brain.
- **D-W2.6 — catch-alls retired structurally**: both `_ => EffectSet.Empty` sites replaced with exhaustive node-kind switches; unrecognized nodes → Calor0419, never silently pure. Pinned with a synthetic unknown node.

## New diagnostics
Calor0418–0421, registered in Diagnostic.cs + docs/syntax-reference/effects.md (self-check clean).

## Behavioral test changes (all justified in the diff)
Two hole-pinning tests now pin the closure; lambda/delegate emission-shape tests carry the explicit waiver; three §USING conversion tests follow the converted-code precedent; two samples gain honest `mut` declarations; E2E agent-task harness gains per-task `permissiveEffects` (lambdas-delegates 01/02/04).

## Tests
+18 new pins (StrictnessBatchTests 16, CLI default 2 via real subprocess). Full suite: **0 failed / 7,835 passed / 17 skipped**, 0 warnings, self-check docs clean, calor-first guard clean.

## Scope-guard note
The A-1.2/A-1.3 frozen exclusion rows cite the delegate-invocation and dispatch-laundering holes; both are now closed. Additive note to be recorded against the frozen rows (gates doc untouched by this PR).

### DIFF
--- a/docs/cli/compile.md
+++ b/docs/cli/compile.md
@@ -55,6 +55,7 @@ calor -v -i MyModule.calr -o MyModule.g.cs
 | `--strict-api` | | No | Require `§BR` breaking-change markers for public API changes |
 | `--require-docs` | | No | Require documentation on public functions |
 | `--enforce-effects` | | No | Enforce effect declarations (default: true) |
+| `--no-enforce-effects` | | No | Disable effect enforcement (opt out of the v0.11 default-on behavior) |
 
 ---
 
--- a/docs/syntax-reference/effects.md
+++ b/docs/syntax-reference/effects.md
@@ -283,12 +283,25 @@ You cannot hide an effect by burying it in helper functions.
   §C{f001:Helper}                  // ERROR: cw effect leaks through
 ```
 
+### Soundness Diagnostics (v0.11 strictness batch)
+
+Effect enforcement is fail-closed. The diagnostics on the enforcement surface:
+
+- **Calor0410** — a function uses an effect it does not declare (with call chain).
+- **Calor0411** — an unknown external call; add the callee to a `.calor-effects.json` manifest. Unknown calls contribute worst-case effects, so under the default policy they fail loud rather than being assumed pure.
+- **Calor0418** — invocation of a delegate/function-typed value (a parameter, binding, or field being called). Function-typed values carry no effect contract, so the call is an **error** under enforcement. There is no annotation escape hatch; wrap the call in `§CSHARP` interop (surfaced as an assumption via Calor0419) or compile with `--permissive-effects` (an explicit waiver that demotes the error to a warning).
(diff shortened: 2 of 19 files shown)


juanmicrosoft:
**Adversarial review record** — Round 1: NEEDS-FIXES (5 C + 1 M: §PP silently pure both legs, decoy-name shadowing evading Calor0418, inherited-implementation interface laundering, method-group laundering through kept HOF names, dotted-name collision pre-empting the fail-loud chain, 0418 demoted by ExpressionCall respelling). Fixes at f028b90b. **Verification round: MERGE-CLEAN** — all 8 finding repros now fail loud with the claimed diagnostics, all 10 legit-behavior repros unregressed, 6/6 fresh evasion attempts failed loud (nested-block shadow, §PP else-only, §PP-in-§PP, two-level inheritance chain, internal-HOF method group, declared-but-external-receiver collision), full suite 7,846/0, samples green under flipped default, effects.md soundness sentence verified accurate. One disclosed residual (method-group rebound to a local then passed to a known-pure HOF → Calor0419 disclosure, not silent) and one cosmetic residual (id-first §FLD type string in the 0418 message) accepted.

PR #843, opened 2026-08-02
docs: WS-W4 kickoff — real-scale benchmark prerequisites + results-blind decisions

juanmicrosoft:
The W4 analogue of `wedge-w1-prereqs.md`. Records the results-blind decisions made at W4 kickoff; **freezes nothing** — A-1.4 tranche 2 freezes after the D-W4.4 dry-run per the split clock (wedge plan §2 D-W4.5, review n3).

## What this decides (results-blind, at kickoff)
- **Governing principle** (§1): predicate trust = *"no silent substitution survives native conversion"* — satisfiable by **loud-refusal** (interop escalation / diagnostic / build break → non-native → excluded by predicate clause (a)), **not** full feature migration. This collapses the fidelity-epic triage into a bounded SILENT→LOUD pass.
- **Migration-fidelity go/no-go** (§2): #774 full de-silencing (in flight); #772/#769 minimal SILENT→LOUD refusal (refuse to keep-first-`#if`-branch / refuse to merge same-named types) with full preservation deferred; #775/#777 containment + registry honesty; #751 already loud (`Calor0258`), no predicate fix.
- **D-W4.5 annotation protocol** (§3): provisionally **(iii) both-as-arms** (arm i mechanical-only = day-one product; arm ii authored-under-blindness = proof channel), **dry-run-gated** on (1) deterministic-contract-overlay feasibility and (2) 3-arm cost vs the $1,500 ceiling, with **(i) mechanical-only as the pre-committed fallback**. PP-W2 wording drafted for both outcomes (restate-or-demote).
- **D-W4.1 predicate** mapped onto the harness: `ConvertedNative` (LossCount==0 ∧ not reverted) + identical held-out-test failure on both arms; mutate-then-convert; presentation asymmetry (converted `§`-syntax vs idiomatic C#) recorded as a bias *against* Calor.
- **D-W4.3 fidelity gate** structure (`NativeFraction` bar, reverted-in-denominator anti-masking, attribution via `BisectRegressionsAsync`); threshold **provisional 0.70**, frozen tranche 2. **D-W4.4 ceiling-recurrence floor** with the pre-committed "ceiling persists at real scale" → not-adjudicated branch.

## Prerequisite / logistics
- **Corpus (D-W4.2)**: vendor MediatR/Serilog/FluentValidation at pinned SHAs + Synthetic; end the external un-pinned dependency; baseline-green measured early; ≥2 projects pass or PP-W2 not adjudicated.
[...]

Issue #774, opened 2026-07-24
P1: Eliminate silent expression, operator, assignment, and pattern substitutions


juanmicrosoft:
## Summary

Eliminate silent literal, operator, assignment, and pattern substitutions in the C# converter.

## Evidence and impact

- Character literals become string literals, producing invalid comparisons and changing overload resolution.
- Assignment expressions ignore their actual operator.
- Unknown binary operators default to addition.
- Unknown prefix operators default to negation.
- Unsigned-right-shift operators are not fully represented.
- Typed declaration patterns lose their type test.
- Unknown patterns become wildcards, broadening a switch arm to match everything.

These substitutions are parse-valid often enough to evade syntax-only gates while changing behavior.

## Required implementation

1. Add a dedicated character literal representation that survives both emitters.
2. Replace default operator mappings with exhaustive `SyntaxKind` switches.
3. Represent every compound assignment, including unsigned shifts, distinctly.
4. Add typed declaration and type-only pattern nodes.
5. Prohibit wildcard/add/negate as generic fallbacks.
6. Escalate unsupported expressions and patterns to containing-member interop in lossless mode.
7. Audit numeric suffix, width, signedness, and overload-resolution preservation.
8. Generate a registry test covering every relevant Roslyn operator/pattern kind.

## Required regression coverage

- Char comparisons, switch cases, patterns, arrays, APIs, overloads, and generic inference.
- Every binary, unary, and compound-assignment `SyntaxKind`.
- Assignment expressions in returns, arguments, chains, and conditions.
- `>>>` and `>>>=`.
- Declaration, type, property, positional, recursive, and list patterns.
- Unsupported pattern proving it cannot broaden match behavior.

## Success metrics / definition of done

- Exhaustive mappings contain no semantic default branch.
- Every supported literal retains its runtime type and overload behavior.
- Pattern matrices produce the same selected arm and bindings as source C#.
- Unsupported cases cannot return successful native conversion.
- Existing char/operator/pattern fixtures compile and pass runtime equivalence tests.

## Dependencies

Depends on semantic validation and the lossless interop contract.

juanmicrosoft:
[...]

Issue #772, opened 2026-07-24
P0: Preserve conditional-compilation semantics by default


juanmicrosoft:
## Summary

Preserve conditional-compilation and compiler-directive semantics by default. The converter must not choose branches by source order or silently delete compiler-affecting directives.

## Evidence and impact

- `StripPreprocessor` defaults to `true` and runs before Roslyn parsing.
- The stripper always keeps the first `#if` branch, without evaluating its condition or symbols.
- `#nullable`, `#pragma`, `#warning`, `#error`, and `#line` are deleted.
- Project migration inherits the unsafe default.
- The support registry nevertheless describes preprocessor directives as fully supported.

`#if false ... #else ...` can therefore convert dead code and discard live code.

## Required implementation

1. Change the default to preserve preprocessors.
2. Preserve all branches using explicit preprocessor AST blocks in lossless conversion.
3. Accept project parse options and defined symbols from MSBuild/project migration.
4. If branch selection remains available, rename it to make loss explicit and require opt-in.
5. Evaluate conditions through Roslyn using the provided symbol set.
6. Preserve nonconditional directives or emit explicit lossy diagnostics for each removal.
7. Record the selected configuration/symbol set in conversion metadata.
8. Downgrade the feature registry until the behavior is covered.

## Required regression coverage

- `#if false/#else`, defined and undefined symbols, and nested `#if/#elif/#else`.
- Conditional usings, members, statements, and partial declarations.
- `#nullable`, warning suppression, `#warning`, `#error`, and `#line`.
- Project migration under multiple configurations/symbol sets.
- Lossless versus explicitly selected-branch behavior.

## Success metrics / definition of done

- Default conversion never selects a branch by source position.
- All conditional branches survive lossless conversion.
- Symbol-selected output matches Roslyn’s active syntax for the same parse options.
- Every removed directive is explicitly reported as loss.
- The production default is exercised by all preprocessor tests.
- The support registry claim matches the implemented subfeature matrix.

## Dependencies

Depends on the lossless conversion contract and semantic validation gates.

PR #844, opened 2026-08-02
fix: eliminate remaining silent conversion substitutions (#774) — W4 Slice A keystone

juanmicrosoft:
Closes the remaining silent-substitution gaps in the C#→Calor converter (#774). The operator/pattern/assignment/char **core** was largely delivered by W1 Slice 3 (a326fd5e); this PR closes the residual gaps and adds the missing type-test-preserving pattern node + no-silent-default proof tests. This is the **keystone of WS-W4 Slice A** (conversion-honesty prerequisite): per the W4 kickoff (#843), the eligibility predicate is only trustworthy if no silent substitution survives native conversion.

## Fixes
- **Prefix-unary escalation**: `ConvertPrefixUnaryExpression` had `?? UnaryOperator.Negate` — any unknown prefix op (including unary `+x`) silently became `-x`. Replaced with an exhaustive `SyntaxKind` switch that escalates unknowns to interop.
- **`TypePatternNode` added** (full AST checklist): `case Type name:` / `case Type:` previously collapsed to a bare `VarPatternNode`/`LiteralPatternNode` that **matched every value and broadened the arm**. Now the runtime type test is preserved (`§PTYPE`).
- **No generic fallbacks**: replaced silent `_ => "+"`/`"add"`/`"+="`/`"_"` defaults in both emitters with fail-loud `ArgumentOutOfRangeException`; list-pattern slice carrying an inner sub-pattern (`..[1,2]`) now escalates instead of dropping the constraint to `_`.
- **Registry conformance test**: a no-silent-default `[Theory]` over all 18 binary operators (each maps to its own Calor operator, never `(+ a b)`), plus prefix-unary escalation, type/declaration-pattern preservation, and generic `§PTYPE{Box<int>:b}` round-trip.

## Deliberately escalated to interop (not modeled), with why
Unary `+x` (no Calor node; could change `operator+` overload resolution); `>>>`/`>>>=` (pre-existing deliberate escalation — no unsigned-shift in Calor's operator enums; existing merged tests assert it); list-slice inner sub-patterns; any unknown binary/prefix/pattern operator.

## Goldens updated (verified correct)
`13-04`/`13-05.approved.calr` — old goldens **encoded the bug** (`§VAR{ok}`/`§VAR{n}`), now correctly `§PTYPE`; `SwitchExpressionConversionTests.Convert_SwitchExpression_VarPattern` asserts `TypePatternNode` instead of the type-dropping `VarPatternNode`. CRLF preserved.

## Entangled siblings found (NOT fixed here — flagged for the Slice A structural batch)
- **Numeric literal-suffix fidelity**: unsigned literals fitting in `long` (`4000000000u`) and `float` literals (`3.14f`) lose their suffix (Calor's numeric surface has no unsigned/float marker) → **silent** overload-resolution/precision change. Same family as #774; needs an unsigned/width tag or loud escalation. Belongs to the #772/#769/#775/#777 + numeric-loudness Slice A batch.
- `ConvertPatternToExpression` is-expression path drops a declaration pattern's *binding* (preserves the type test) — pattern-completeness, not arm-broadening.

## Tests
Full suite green (~7,875 passing, 0 failing, 11 projects); Release build clean; `calor self-check docs` no drift.

### DIFF
--- a/src/Calor.Compiler/Ast/AstNode.cs
+++ b/src/Calor.Compiler/Ast/AstNode.cs
@@ -164,6 +164,7 @@ public interface IAstVisitor
     void Visit(NegatedPatternNode node);
     void Visit(OrPatternNode node);
     void Visit(AndPatternNode node);
+    void Visit(TypePatternNode node);
     // Extended Features Phase 1: Quick Wins
     void Visit(ExampleNode node);
     void Visit(IssueNode node);
@@ -388,6 +389,7 @@ public interface IAstVisitor<T>
     T Visit(NegatedPatternNode node);
     T Visit(OrPatternNode node);
     T Visit(AndPatternNode node);
+    T Visit(TypePatternNode node);
     // Extended Features Phase 1: Quick Wins
     T Visit(ExampleNode node);
     T Visit(IssueNode node);
--- a/src/Calor.Compiler/Ast/ExpressionNodes.cs
+++ b/src/Calor.Compiler/Ast/ExpressionNodes.cs
@@ -21,6 +21,15 @@ public sealed class IntLiteralNode : ExpressionNode
     public bool IsUnsigned { get; }
     public ulong UnsignedValue { get; }
 
+    /// <summary>
+    /// True when the literal carries an explicit 64-bit width (C# <c>L</c>/<c>UL</c>
+    /// suffix, or a <c>ulong</c>). #774: preserved so a <c>long</c>/<c>ulong</c>
+    /// literal that still fits in a smaller type is not silently narrowed and its
+    /// overload resolution is kept. When false, the emitter derives width from
+    /// magnitude as before.
+    /// </summary>
+    public bool IsLong { get; init; }
+
     public IntLiteralNode(TextSpan span, long value) : base(span)
     {
         Value = value;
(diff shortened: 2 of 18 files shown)


juanmicrosoft:
**Adversarial review record** — Round 1: NEEDS-FIXES (1 C, block-now). Everything the PR touched (patterns/operators/char/unary/§PTYPE type-test preservation) verified CLEAN under empirical attack; blocker C1 = numeric literal suffix loss surviving native conversion (float→double, uint→int/long, long→int — silent overload/precision/signedness change). **Fixed by faithful preservation** at 5281c56c: new SINGLE:/LONG:/UINT:/ULONG: typed literals mirroring the DECIMAL: mechanism — no interop escalation, benchmark coverage preserved. **Verification round: MERGE-CLEAN** — all six repros round-trip faithfully in both var and overload contexts (3.14f→3.14f float exact, 7u→uint not int, 4000000000u→uint not long, 5L→long); new literal surface attacked for precision drift / lexer collision / magnitude-vs-marker precedence / edge cases (0xFFL, 1_000L, 1e10f, uint.MaxValue, ulong.MaxValue) — all correct; Z3 inert (markers metadata over .Value); suite ~7,884/0. C1 genuinely closed: no numeric silent substitution survives native conversion. Deferred (non-blocking): m1 is-expression binding drop (loud, not silent); a PRE-EXISTING Z3 unsigned-obligation modeling limitation (UINT:/ULONG: via signed .Value) tracked as a separate verification-semantics follow-up.

Issue #777, opened 2026-07-24
P0: Preserve local-function scope, closure, recursion, and generics


juanmicrosoft:
## Summary

Preserve local-function scope, closure captures, recursion, generics, and containing-instance context during migration.

## Evidence and impact

- Local functions are hoisted into the module function collection.
- Type parameters are discarded.
- Call sites remain in the containing class/method and may no longer resolve.
- Captured locals, `this`, recursion, async/static/unsafe modifiers, and ref semantics are not represented.
- Existing tests acknowledge out-of-scope captures while still asserting conversion success.

### Immediate containment

- Preserve any method containing local functions as member-level interop in lossless mode.

### Native support

1. Add a nested-function representation, or implement a formally specified closure conversion.
2. Preserve immutable and mutable captures, containing `this`, capture lifetime, and ref/out/scoped semantics.
3. Preserve recursion and mutually recursive local functions.
4. Preserve type parameters, constraints, overloads, attributes, `static`, `async`, and `unsafe`.
5. Ensure generated helper calls are qualified and collision-free if lifting is used.
6. Reject or interop unsupported capture forms rather than emitting free variables.

## Required regression coverage

- Noncapturing, capturing, and mutating local functions.
- Instance-member capture.
- Recursive and mutually recursive functions.
- Generic functions with constraints.
- Static, async, unsafe, iterator, and overloaded local functions.
- Ref/out/scoped parameters and captured variables.

## Success metrics / definition of done

- Every supported local-function fixture compiles and passes behavioral tests.
- No generated helper has an unresolved free variable.
- Capture mutation/lifetime behavior matches the source.
- Generic constraints and modifiers survive syntax/API comparison.
- Unsupported forms use whole-member interop.
- Existing local-function false-positive scorecard entries are corrected.

## Dependencies

Depends on semantic round-trip gates and the lossless interop contract.

Issue #775, opened 2026-07-24
P0: Implement faithful record migration or preserve records as interop


juanmicrosoft:
## Summary

Implement faithful C# record semantics or preserve records as interop. Stop classifying the current sealed-class downgrade as full support.

## Evidence and impact

- The support registry marks records fully supported.
- Positional parameters become getter-only properties.
- No corresponding positional constructor is generated.
- Output is always a sealed class.
- Record class/struct distinction, equality/hash, copy construction, `with`, `Deconstruct`, base record/interfaces, attributes, and several explicit members are lost.
- Existing fixtures emit `new Point(...)` against a generated type with no matching constructor and are still marked fully converted.

### Immediate containment

- Downgrade record support.
- Preserve records or the containing type as interop in lossless mode.

### Native support

1. Introduce a dedicated record AST retaining record class versus record struct, positional parameters, attributes, modifiers, generics/constraints, base record/interfaces, and explicit members.
2. Prefer emitting native C# record syntax on round trip.
3. Preserve synthesized behavior: constructor, equality/hash, copy/clone, `with`, deconstruction, and printable representation.
4. Preserve user-defined equality, copy constructors, and explicit property/member overrides.
5. Promote support per record subfeature only after behavioral validation.

## Required regression coverage

- Positional and nominal records.
- Constructor invocation and named/positional parameters.
- Equality and hash behavior.
- `with`, `Deconstruct`, and derived records.
- Record structs and readonly record structs.
- Generic records, constraints, interfaces, and attributes.
- Public API signature comparison.

## Success metrics / definition of done

- Every variant marked supported compiles and passes behavioral equivalence tests.
- Constructor, equality/hash, `with`, and deconstruction results match source behavior.
- Public API comparison reports no signature loss.
- Unsupported variants remain verbatim interop and are never counted as fully native.
- Existing false-positive record scorecard entries are corrected.

## Dependencies

Depends on semantic round-trip gates and the lossless interop contract.

Issue #770, opened 2026-07-24
P0: Define and enforce a genuinely lossless conversion contract


juanmicrosoft:
## Summary

Define and enforce a genuinely lossless conversion/interop contract. Unsupported descendants must never become silent TODOs, wildcards, dropped members, or successful output.

## Evidence and impact

- Fallback expressions become parse-valid `§ERR "TODO..."` regardless of interop mode.
- Unsupported statements do not consistently honor `Mode == Interop`.
- Unsupported class/interface members can be dropped.
- Standard mode can return success without validating emitted Calor.
- CLI `--validate` writes output before completing validation and still reports success for some failures.

The current behavior makes it impossible for callers to distinguish faithful conversion from semantic loss.

## Required implementation

1. Formalize two explicit modes:
   - **Lossless:** preserve unsupported code at the nearest complete member/type boundary.
   - **Lossy:** require explicit caller opt-in and report every semantic loss.
2. Apply interop semantics consistently to expressions, statements, members, types, and whole files.
3. Bubble an unsupported child to a containing compilable-member interop fallback.
4. Validate emitted Calor in every mode.
5. Validate round-tripped C# compilation in lossless mode.
6. Make unacknowledged drops/substitutions set `Success=false`.
7. Write destinations atomically only after requested validation succeeds.
8. Add structured counts and locations for native conversion, interop preservation, lossy substitutions, and drops.
9. Document the guarantee and expose it consistently through CLI, SDK, MCP, and project migration.

## Required regression coverage

- Unsupported expression nested in an otherwise supported method.
- Unsupported statement, interface member, class member, attribute, and modifier.
- Parse-valid fallback expression.
- Validation failure with an existing destination file.
- Explicit lossy opt-in and exact loss locations.
- Lossless project conversion with mixed supported/unsupported members.

## Success metrics / definition of done

- Lossless mode has zero silently dropped nodes.
- Every unsupported construct is faithfully represented or preserved verbatim.
- Every successful lossless result parses as Calor and compiles as C#.
- Validation failure exits nonzero and leaves the destination byte-for-byte unchanged.
- Any semantic loss produces nonzero structured loss counts.
- All public conversion surfaces use the same mode semantics and result schema.

## Dependencies

Depends on semantic round-trip validation. Other migration issues should use this as immediate containment via member/type interop.

juanmicrosoft:
[...]

Issue #769, opened 2026-07-24
P0: Preserve namespace topology and namespace-local using scopes


juanmicrosoft:
## Summary

Preserve C# namespace topology, fully qualified type identity, and namespace-local using scopes during conversion.

## Evidence and impact

- A module name is derived from only the first namespace.
- Namespace visitors mainly update diagnostic context; discovered types are accumulated into one module.
- Namespace-local usings are added to one global collection.
- Post-validation acknowledges that the module flattens namespaces and recovers members by bare kind/name.

This changes type identity, introduces collisions for same-named types, leaks aliases/usings, and can merge unrelated partial types.

## Required implementation

1. Add namespace hierarchy/scope to the Calor AST, or emit separate modules per namespace with an explicit mapping.
2. Associate every type, delegate, and top-level member with its namespace identity.
3. Preserve namespace-scoped aliases, static usings, and global usings.
4. Key source mapping, recovery, and partial-type merging by fully qualified symbol identity.
5. Update project migration output naming to avoid cross-namespace collisions.
6. Until native support is complete, preserve multi-namespace files as whole-file interop in lossless mode.

## Required regression coverage

- Multiple namespaces in one file.
- Same type name in different namespaces.
- Nested and file-scoped namespaces.
- Conflicting namespace-local aliases.
- Global using versus namespace-local using.
- Partial types across files and namespaces.
- Project conversion output-path collisions.

## Success metrics / definition of done

- Round-tripped fully qualified type identities match the source.
- Same-named types in different namespaces compile without collision.
- Namespace-local using visibility never leaks.
- Recovery and partial merging use symbol identity, not bare names.
- No output file is overwritten because two namespace/type identities map to one path.
- Unsupported topology is preserved by interop rather than flattened.

## Dependencies

Depends on the lossless contract and semantic validation.

PR #846, opened 2026-08-02
feat: WS-W4 Slice A structural — SILENT→LOUD for namespace/local-functions (+ pin preprocessor/records)

juanmicrosoft:
The structural half of WS-W4 Slice A (conversion-honesty prerequisite). Governing principle (merged kickoff #843 §1): no silent substitution may survive *native* conversion — unfaithful conversions must be LOUD (interop loss / build-break) so the eligibility predicate excludes them. Empirical probing found D1/D3 were already loud from #770/#773; the new work is D2 and D4.

## Deliverables
- **D1 — #772 preprocessor (already loud; pinned).** The default pipeline records a `PreprocessorStripped` loss for every `#if`/`#elif`/`#else` → non-native. Verified the `#if false…#else…` dead/live swap is non-native. Registry already honestly `Partial`. Added 3 pins.
- **D2 — #769 same-name namespace merge (NEW loud refusal).** `ScanCrossNamespaceCollisions` (full namespace path) + `TryRefuseCrossNamespaceCollision` in class/struct/interface/enum/delegate visitors: two top-level types sharing a bare name across ≥2 namespaces both escalate to `§CSHARP` interop (`namespace-collision` loss) instead of silently flattening to one identity. Unique-name flattening stays native (round-trip verified); same-name partials in one namespace remain a legitimate merge. Registry `namespace` = Partial.
- **D3 — #775 records (already loud; pinned).** Records escalate verbatim to interop (`record` loss; registry `NotSupported`). Added a `with`/`==` value-semantics regression pin.
- **D4 — #777 local functions (NEW loud refusal).** Generic or capturing local functions escalate the containing member to interop, detected at the member-body boundary in `ConvertBlock` before any hoist (no half-hoist duplication). Capture detection over-approximates (false positive → extra interop, never a wrong native). Non-generic non-capturing local functions (incl. recursion) stay on the hoist path. Registry `local-function` = Partial.

## Honest channel
No new `Calor####` codes — the converter's honest channel is the structured loss ledger (`InteropPreserved` losses tagged `namespace-collision`/`record`/`local-function`/`generic-delegate`) which the predicate counts.

## Flagged gaps (see PR discussion for the load-bearing one)
1. **Plain local functions are hoisted to a module static class while the call site stays in the origin class → CS0103 build-break at round-trip** (LOUD → revert → non-native, per the agent). **This PR's review must verify this is ALWAYS loud and cannot silently rebind** to a same-named origin-class member. Pre-existing emitter limitation; faithful fix (emit real C# local functions) is a larger out-of-scope emitter change.
2. Instance/`this`/field capture stays native and build-breaks (loud, per task acceptance).
3. Top-level-statement local functions bypass the member-body scan (narrow — the pinned OSS corpus are libraries).
4. D1 standalone `#nullable`/`#pragma`/etc. deleted without a loss (deferred; non-runtime-semantic; active `#error` excluded by baseline-green).
5. D2 unique-name flattening changes fully-qualified identity (task-permitted; registry no longer claims faithful namespace preservation).

## Tests
Full suite green (13 projects; Conversion 364, Compiler 6148, Verification 347, Enforcement 301, …); Release 0 warnings; self-check docs no drift. One pre-existing cross-namespace-collision test updated to the new cleaner behavior.

### DIFF
--- a/src/Calor.Compiler/Migration/FeatureSupport.cs
+++ b/src/Calor.Compiler/Migration/FeatureSupport.cs
@@ -577,6 +577,35 @@ public static class FeatureSupport
             Workaround = "Review conditional-compilation blocks manually; alternate branches are not preserved in the default pipeline"
         },
 
+        // #769 (WS-W4 D2): namespace topology is NOT preserved — every type in a
+        // file is flattened into a single module named after the first namespace,
+        // so a type's fully-qualified identity changes. Flattening that carries no
+        // identity-merge risk (single namespace, or types with unique bare names)
+        // stays native; two top-level types that share a bare name across different
+        // namespaces would collapse to one identity, so they are refused and escalated
+        // to §CSHARP interop (a counted loss) rather than silently merged.
+        ["namespace"] = new FeatureInfo
+        {
+            Name = "namespace",
+            Support = SupportLevel.Partial,
+            Description = "Namespace topology is flattened into one module; cross-namespace same-name types are refused (escalated to §CSHARP interop) rather than silently merged",
+            Workaround = "Split multi-namespace files, or keep same-named cross-namespace types as §CSHARP interop blocks"
+        },
+
+        // #777 (WS-W4 D4): local functions are lowered by hoisting to module-level
+        // §F functions. A non-generic, non-capturing local function round-trips
+        // faithfully. A generic local function would lose its type parameters, and a
+        // capturing local function would reference enclosing locals that do not exist
+        // at module scope — both are refused and escalate the containing member to
+        // §CSHARP interop (a counted loss) rather than silently diverging.
+        ["local-function"] = new FeatureInfo
+        {
+            Name = "local-function",
+            Support = SupportLevel.Partial,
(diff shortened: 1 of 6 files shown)


juanmicrosoft:
**Adversarial review record** — Round 1: NEEDS-FIXES (3 C + 1 M). The load-bearing gap #1 was empirically exploitable: plain local functions silently rebound to same-named class/base/overload members and converted natively (C-1/C-2/C-3 — LossCount==0, compiles, wrong behavior). Plus M-1 (D2 falsely merged Foo vs Foo<T> on arity-blind keying). **Fixed at 12104094**: escalate ALL local functions to §CSHARP interop (the hoist path was proven already non-native on its happy path via CS0103, so escalate-all loses no legitimate coverage while closing the silent path); D2 collision key now includes generic arity. Registry local-function → NotSupported. Follow-up #847 filed for faithful in-place local-fn emission (coverage enhancement). **Verification round: MERGE-CLEAN** — C-1/2/3 now non-native (method→InteropPreserved); swept every bypass shape (nested locals, locals in lambdas/accessors/try/switch/using/ctor/operator/expression-bodied, top-level statements) — none survives native; no false over-escalation; arity keying correct both directions; ~8,349 tests/0 failures. Slice A (converter honesty) complete.

PR #852, opened 2026-08-03
feat: WS-W4 — complete the gold-standard bug-fix-revert task source

juanmicrosoft:
Completes and wires the revert-upstream-bugfix task source (the gold-standard, real-defect source the Slice-C `BugfixMiner` stubbed). A real-corpus run proved the stub yielded 0 live tasks; this fixes the mining path and reports the honest yield.

## Fixes
- **Pathspec bug**: `GitLogArgs` no longer restricts `git log` to the source prefix (which meant `--name-status` never listed test files → `SelectRevertCandidates` could never fire). Now surfaces both source + test changes; classifier tightened to `prefix + "/"` so sibling test dirs aren't miscounted. Live: **278 revert candidates** (MediatR 9, Serilog 127, FV 142).
- **Source-hunk-only revert**: reverse-applies ONLY the fix commit's library-source hunks (`git apply -R --3way` in a throwaway worktree), keeping the covering test at its fixed state so it becomes the held-out oracle (a whole-commit revert would delete it). Loud on conflict/no-op/multi-source → `InseparableRevert`/`MultipleSourceFiles`, counted + disclosed.
- Two bugs found+fixed validating against the real corpus: silent no-op `git apply -R` (→ `--3way` + reverted≠pinned guard); **CRLF corruption** in git-stdout capture that made ALL FluentValidation reverts falsely inseparable (→ byte-exact read; FV clean reverts 0→3).
- **Wired into `gen-tasks`**: `--source injected|revert|both` (default injected), same eligibility predicate + exclusion accounting; added a cheap `mine` diagnostic command.

## Honest live yield
278 candidates → 17 cleanly-separable single-source reverts (MediatR 1, Serilog 13, FV 3) → **0 eligible through the strict native-region ∩ compiles ∩ survives gates**. Genuine attrition, not breakage (Serilog's `LevelAlias.cs` matched native then correctly hit `DidNotCompile` — a feature-add later code depends on). The bottleneck is {separable single-source revert} ∩ {natively-converting file}: fix-touched files skew complex/interop-heavy. Injected mutations (23 eligible) remain the workable substrate; this source is correct infra + honest measurement, growable later via multi-source reverts (+135 candidates) or corpus expansion.

## Tests
`TaskGenRevertTests` (synthetic git fixtures, corpus-independent): pathspec fix, clean revert, far-away 3-way, adjacent-conflict→Inseparable, CRLF regression, multi-source, whole-commit-would-delete-the-test property. Harness 113 tests green; 0 warnings.

### DIFF
--- a/tests/Calor.RoundTrip.Harness.Tests/TaskGenGateTests.cs
+++ b/tests/Calor.RoundTrip.Harness.Tests/TaskGenGateTests.cs
@@ -98,6 +98,20 @@ public class TaskGenGateTests
         Assert.Equal(1.0 / 9.0, acc.EligibilityRate, 3);
     }
 
+    [Fact]
+    public void ExclusionAccounting_CountsRevertSupplyExclusions()
+    {
+        var acc = new ExclusionAccounting();
+        acc.Record(Disp(false, ExclusionReason.MultipleSourceFiles));
+        acc.Record(Disp(false, ExclusionReason.InseparableRevert));
+        acc.Record(Disp(false, ExclusionReason.InseparableRevert));
+
+        Assert.Equal(1, acc.ExcludedMultipleSourceFiles);
+        Assert.Equal(2, acc.ExcludedInseparableRevert);
+        Assert.Equal(0, acc.Eligible);
+        Assert.Equal(3, acc.Considered);
+    }
+
     private static CandidateDisposition Disp(bool eligible, ExclusionReason reason) => new()
     {
         ProjectName = "P",
--- a/tests/Calor.RoundTrip.Harness.Tests/TaskGenMinerTests.cs
+++ b/tests/Calor.RoundTrip.Harness.Tests/TaskGenMinerTests.cs
@@ -65,11 +65,16 @@ public class TaskGenMinerTests
     }
 
     [Fact]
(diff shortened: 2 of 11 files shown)


juanmicrosoft:
**Adversarial review record** — MERGE-CLEAN on the git-manipulation core (source-hunk-only revert keeps the covering test as oracle; CRLF + no-op-revert bugs fixed; --3way conflict guard; predicate reuse with no gate bypass; 0-yield verified genuine supply-starvation not breakage). One M (honesty): `IsFixShaped` counted every `#NNNN` PR, overstating 'gold-standard bug-fix' supply. **Fixed at 29f040fb**: require fix-vocabulary (dropped the bare issue-ref regex); corrected counts 278→136 revert candidates, 17→5 clean single-source reverts, and clean reverts now surface genuine logic fixes (e.g. Serilog cf2c62e3 'malformed JSON when renderMessage=true'). Test pins bare-`#NNNN` NOT fix-shaped. Correct gold-standard infra with honest low yield (0 eligible through native∩compiles today; growable via multi-source reverts / corpus expansion / higher converter fidelity). 118 harness tests green.

PR #853, opened 2026-08-03
feat: WS-W4 real-scale epoch runner (bundle dry-run runner) — null-agent verified

juanmicrosoft:
The epoch runner that executes the WS-W4 dry-run: runs an agent per (task bundle, arm, run) and adjudicates catch-vs-escaped via a silent held-out oracle. An adapter over the battle-tested `run-pair.sh` (reuses its invalid-retry, watchdog, `dotnet` PATH shim / silent held-out run, cost capture, `--null-agent`). Five new files under `bench/phase0-agent-native/` (guard-exempt). **No API budget spent; no real agent spawned — verified entirely with `--null-agent`.**

## Adjudication
`caught` = held-out PASSES ∧ visible green; `escaped` = held-out FAILS or final build fails; `broke-regression` = held-out passes but a previously-green visible test regressed; `invalid` = retried then recorded. Cost = summed `total_cost_usd` (never hand-priced); output tokens, iterations, iterations-to-declared-done, wall-clock captured per run. `analyze-bundle-epoch.py` emits the D-W4.4 ceiling-recurrence signal (C#-arm escaped incidence).

## Arms (D-W4.5 provisional-i, mechanical-only)
`csharp` = idiomatic C# + compiler + visible tests; `calor` = round-tripped Calor working copy + enforcement/import annotations (no authored §Q/§S). Both get the identical scrubbed failing-behavior report + identical visible/held-out split. Presentation asymmetry (calor works on machine-converted C#) recorded in every result.json — a bias against Calor.

## Oracle integrity (the load-bearing property)
Two-copy architecture: the **agent-visible** copy has each held-out method PHYSICALLY STRIPPED (brace-matched, string/comment-aware C# removal, incl. `[Theory]` rows + attributes), so the agent cannot read or run the oracle; the **oracle** copy (isolated mktemp tree the agent has no path to) retains it and is kept in sync with the agent's LIBRARY edits (test dir excluded). Fail-loud if any held-out method isn't found / is ambiguous / survives removal (a silent miss = a fabricated measurement → run aborts).

## Verification (zero spend)
- Oracle-leak proof (FluentValidation): held-out text in agent copy = 0, oracle copy = 1; agent `dotnet test` = 864/865 green with held-out NOT run; full/oracle = 1 failed (the held-out).
- Adjudication sound both arms/both projects: FLUENT csharp/calor CAUGHT+ESCAPED; SERILOG csharp CAUGHT (8-method held-out cluster stripped) + ESCAPED. Null-agent cost 0.

## Notes for the dry-run
Serilog needs `--max-candidates 40` to reach eligible logic bundles (default 6 → 0). Command to launch (post-spend-auth) in README-bundle-runner.md.

### DIFF
--- a/bench/phase0-agent-native/README-bundle-runner.md
+++ b/bench/phase0-agent-native/README-bundle-runner.md
@@ -31,20 +31,32 @@ cost/token/iteration extraction.
 
 **Added (what a bundle forces):**
 
-1. **Whole-project workspace.** `materialize` copies the bundle's `csharp-arm/`
-   or `calor-arm/` (a full OSS project working copy — both are plain `.cs`; the
-   calor arm is round-tripped C#) into `$ws/src`. No synthesized `Src.csproj`:
-   the project builds itself through its own test project (the regression net),
-   using the Slice-B build knobs (`TargetFramework` + `ExtraBuildProperties`)
-   keyed by `ProjectName` (mirrors `ProjectConfigs.cs`). The corpus's `global.json`
-   SDK pin is already dropped in the bundle copy, so it builds on the pinned SDK.
-2. **OSS-project held-out oracle** (`run_oracle`). Builds the regression-net
-   project (rebuilding the mutated library), then runs the **held-out filter**
-   (must PASS = defect fixed) and the **visible filter** (must stay green = no
-   regression). Filter-based split per Slice-C: `HeldOut[].FilterName` →
-   `FullyQualifiedName~…`; `VisibleTestFilter` is the `!~` complement. Run
-   silently by the shim (held-out leg, for the iterations-to-green signal) and at
-   declared-done (both legs, for adjudication).
+1. **Two-copy whole-project workspace (the oracle-leak fix).** `materialize`
+   makes TWO copies of the bundle arm (`csharp-arm/` or `calor-arm/` — both plain
+   `.cs`; the calor arm is round-tripped C#):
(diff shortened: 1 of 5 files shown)


juanmicrosoft:
**Adversarial review record** — R1 NEEDS-FIXES: C (agent-facing `dotnet` shim embedded the oracle tree path + held-out filter in plaintext → agent could `cat $(command -v dotnet)` and fabricate CAUGHT, defeating the two-copy strip); M1 (substring test-matching mis-partitioned prefix-sibling held-out — Serilog cand10 ran 10 not 8, shrinking the regression net); M2 (expr-bodied strip not brace-aware → silent file corruption, no fail-loud). Fixed at 6763b9af: oracle run moved entirely to the parent harness (shim journals metadata only); exact `FullyQualifiedName=`/`!=` split (Theory-row-safe, verified); brace/string-aware expr-body scan + fail-loud; minors (underivable-ref→skipped, nonzero-exit→invalid, sync excludes all test roots, integer-guarded parse). **Verification round: MERGE-CLEAN** — reviewer's OWN grep proves oracle path + held-out names absent across shim/src/spec/final-src and the oracle mktemp name (0 everywhere); held-out leg runs exactly 8 with siblings back in the regression bucket (803); expr-bodied fail-loud confirmed; null path cannot spend, cost summed from total_cost_usd. Documented residual: same-UID /proc introspection (path/toolchain isolation, not a sandbox; per-UID/container is the noted follow-up). Runner safe to spend on.

Issue #761, opened 2026-07-24
P0: Make successful compilation guarantee Roslyn-valid generated C#


juanmicrosoft:
## Summary

A default Calor compilation can report success while returning C# that does not compile. Calor type checking is optional and generated output is not validated as a complete Roslyn compilation.

## Evidence and impact

- The normal CLI constructs `CompilationOptions` without enabling `EnableTypeChecking`: `src/Calor.Compiler/Program.cs`.
- Type checking is conditional and defaults off.
- Generated C# is accepted without a Roslyn compilation.
- `CompilationResult.HasErrors` reflects Calor diagnostics only.
- Code-generation telemetry records success before proving that the output is usable.

This turns emitter defects into downstream C# errors and makes CLI exit codes, build caches, migration scorecards, and telemetry falsely optimistic.

## Required implementation

1. Add a project-aware generated-output validation stage after all inputs are emitted.
2. Parse all generated trees together so cross-file references are resolved correctly.
3. Resolve the same framework, runtime, manifest, and project references used by the invoking surface.
4. Convert Roslyn error diagnostics into Calor diagnostics, using `#line`/source-map information where available.
5. Enable Calor semantic/type checking by default.
6. If syntax-only transpilation is still required, expose it as an explicitly unsafe mode such as `--transpile-only`; it must not use the normal success message.
7. Report telemetry success and populate incremental caches only after generated-output validation succeeds.
8. Extract reusable Roslyn compilation infrastructure rather than duplicating reference resolution in CLI, tests, migration, and self-check.

## Required regression coverage

- Deliberately syntax-valid but type-invalid emitted C#.
- Multi-file compilation with cross-file symbols.
- Calor.Runtime and external assembly references.
- Invalid output from postcondition, iterator, record, using-directive, and interpolation regressions.
- CLI text, JSON, and SARIF diagnostic mapping.
- Exit-code and telemetry assertions.
- Explicit unsafe/transpile-only mode behavior.

## Success metrics / definition of done

- Zero default compilations return `HasErrors == false` when the generated Roslyn compilation contains an error.
- All generated trees from one invocation are validated together.
- Generated-output diagnostics identify the originating `.calr` declaration wherever source mapping exists.
- No cache entry or success telemetry event is written for invalid generated output.
- Every checked-in compile fixture is validated by this production path in CI.
- Documentation clearly distinguishes validated compilation from unsafe transpilation.

## Dependencies

[...]

PR #871, opened 2026-08-05
PP-A1 = PASS (all 9 frozen items) + disclose the format --write release gate

juanmicrosoft:
Adjudicates **PP-A1**, a v0.12.0 release gate orthogonal to Call S, against the list frozen at `wedge-w1-prereqs.md` §3 *before any results existed*.

**All nine items satisfied**, each with cited evidence:

| # | Evidence |
|---|---|
| 1 | `sdk-package-consumer` job (M-A1) green |
| 2 | `publish` job declares `needs: [test, sdk-consumer]` |
| 3 | sha256 manifest + `verify_archive` ×3, verification precedes extraction, fail-closed |
| 4 | `ComputeOptionsHash` consumed at `CompilationDriver.cs:113` |
| 5 | `CALOR_TELEMETRY=1` required; opt-out forces off; anonymizing initializer; documented |
| 6 | W1 Slice 1 `484d6c70`; **D6/D7/D8/D9** dispositions in the divergence table |
| 7 | `format --write` refuses with **Calor1346**; LSP formatting + rename register **only** under `CALOR_LSP_EXPERIMENTAL` |
| 8 | eject degradation table in `adoption-playbook.md` |
| 9 | passes **by its own terms** — the frozen list marks the #761 flip *"not gate-blocking for W1 exit"* |

## A defect found by the audit, fixed here

Item 7's containment is implemented **correctly** — but `docs/cli/format.md` documented `--write` as an ordinary option in three places with no mention of the gate. A reader following the docs would hit `Calor1346` unwarned and reasonably read a deliberate containment as a bug. Now disclosed at all three sites.

## A near-miss, recorded

An intermediate step read the `WithHandler<FormattingHandler>()` / `WithHandler<RenameHandler>()` lines **without their enclosing `if (experimentalWriteHandlers)` block** and concluded item 7 was unmet — that a release blocker had *failed*. It had not. **Grep-level evidence about whether something is gated is unreliable by construction**, because the gate is the enclosing context, not the line.

## Scope

PASS is against the frozen list only; its explicit exclusions stay excluded. **`EnableTypeChecking` is NOT default-on at v0.12.0** — item 9 requires a scheduled slot, not a shipped flip. **PP-W5 is not adjudicated here**: it needs a spend-authorized epoch and is a separate maintainer decision.

### DIFF
--- a/docs/cli/format.md
+++ b/docs/cli/format.md
@@ -28,7 +28,7 @@ The `format` command formats Calor source files according to the canonical Calor
 # Format a single file (output to stdout)
 calor format MyModule.calr
 
-# Format and overwrite the file
+# Format and overwrite the file — DISABLED by the release policy (see below)
 calor format MyModule.calr --write
 
 # Check if files are formatted (for CI)
@@ -53,7 +53,7 @@ calor format MyModule.calr --diff
 | Option | Short | Default | Description |
 |:-------|:------|:--------|:------------|
 | `--check` | `-c` | `false` | Check if files are formatted without modifying (exit 1 if not) |
-| `--write` | `-w` | `false` | Write formatted output back to the file(s) |
+| `--write` | `-w` | `false` | Write formatted output back to the file(s). **Disabled by the release policy** — see below |
 | `--diff` | `-d` | `false` | Show diff of formatting changes |
 | `--verbose` | `-v` | `false` | Enable verbose output |
 | `--heal` | — | `false` | Best-effort source-level repair of files too broken for the AST formatter. **Not semantics-preserving** — see [Heal Mode](#heal-mode) |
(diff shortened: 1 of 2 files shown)


juanmicrosoft:
Superseded by #873, which supersedes this document entirely rather than amending it.

This PR claimed **PP-A1 = PASS on all nine items**. Adversarial review refuted that, and it took four more rounds to settle: item 6 was FAIL because divergence **D4** was a live false-`Proven`-elides vector, then still FAIL after the first fix (D4's no-mode half), then FAIL again on **D3**, and **D12** turned up alongside it. All four are now closed — D4 by refusal (#872), D3 and D12 by demoting string-carried proofs to `Assumed` (#876) — and item 9's flip landed in #877.

Items 4 and 7 also passed here on evidence that did not hold: item 4 cited `CompilationDriver` while the MSBuild task real projects build through was missing the output-content check, and item 7's `Calor1346` containment had no test anywhere in the repo.

#873 carries the corrected adjudication, the fixes, and the record of how each row actually failed. Closing this rather than updating it, since almost nothing in it survived.

Issue #779, opened 2026-07-24
P0: Keep runtime guards until verification meets a soundness gate


juanmicrosoft:
## Summary

Make verification fail-safe immediately: no `Proven` outcome may remove a runtime guard until the verifier, cache, and generated-runtime semantics meet an explicit soundness gate.

This is the containment/coordination issue for the broader verification remediation. Existing issue #755 covers the specific precondition satisfiability-versus-validity defect and should be fixed, not duplicated.

## Evidence and impact

- #755 demonstrates that a satisfiable precondition is labeled `Proven` and its runtime guard is removed.
- Postcondition guards are also removed based on solver models that currently diverge from C# numeric and string semantics.
- Cached `Proven` results can collide across distinct expressions.
- Refinement guard policy is not wired to emission.

The unsafe outcome is not merely a bad diagnostic: generated code can omit checks that users expect to enforce contracts.

## Immediate containment

1. Change `--verify` to diagnostic-only behavior.
2. Retain every runtime precondition, postcondition, refinement, and obligation guard regardless of solver result.
3. Introduce an explicit capability flag for any future proof-based elision; default it off in CLI, SDK, MSBuild, MCP, watch, and tests.
4. Emit a structured warning when verification cannot affect runtime checks.
5. Invalidate existing verification caches when containment ships.

## Permanent soundness gate

Before any elision is reenabled:

1. Define typed proof outcomes that distinguish satisfiable, valid, refuted, unknown, timeout, unsupported, and internal failure.
2. Define which exact outcome/context combinations are eligible for elision.
3. Require a complete semantic-model version and collision-safe cache key.
4. Require solver-versus-generated-runtime differential tests for every supported operation/type.
5. Require an independent call-site proof before eliding a precondition.
6. Keep a runtime fallback for unsupported, timeout, unknown, or failed verification.

## Required regression coverage

- The reproduction in #755.
- False proof cases for integer width/promotion, signedness, and string operations.
- Cache-collision reproduction.
- Timeout, unsupported expression, Z3 initialization failure, and internal exception.
- Every public surface and contract mode.
- Runtime execution proving violating inputs still throw while containment is active.

## Success metrics / definition of done

- With containment active, zero verification outcomes remove a runtime guard.
- The #755 violating-call reproduction throws at runtime.
- All verification statuses are semantically distinct and exhaustively handled.
- Unsupported/timeout/error paths are fail-safe and covered.
[...]

PR #873, opened 2026-08-05
PP-A1 = PASS (all nine items) — fixes for items 2, 3, 4, 7; item 9 delivered by #877

juanmicrosoft:
**Supersedes #871**, which claimed all nine items PASS. Adversarial review refuted item 6 and was right — so this replaces that claim rather than amending it.

> **Merge order: this should land after #872.** Item 6 passes *only* as of that PR.

## What the first adjudication got wrong, and how

**Item 6** — bar: *no known false-`Proven`-elides vector*, known set = the divergence table + T1.

**Divergence D4 was a live one.** A non-ordinal `StringComparison` mode in a contract was translated as **ordinal**, the mode-bearing form stayed whitelisted in `ModeledForms`, and `Z3Verifier` never demoted on the warning — a genuine false `Proven`, which `Proven && !IsVacuous` (D-G1.3) acts on by **eliding the runtime check**. Reproduced end-to-end on a shipped CLI path:

| command | before #872 |
|---|---|
| `calor run` | **throws** `ContractViolationException` — the postcondition is genuinely false |
| `calor run --verify` | prints the value — **check deleted** |

The first adjudication cited four rows that *were* closed (D6–D9) and **never audited the other seven**. This one audits all eleven, and adds:

- **D3** — an explicit *argued-unreachable* disposition rather than silence: Calor's `str` is non-nullable, `?str` is not in `ModeledForms.ScalarTypes` (`NormalizeTypeName`'s default arm keeps the `?`), and with a precondition a `null` argument fails that precondition, which is never elided (D-G1.2). Stated as an argument, not a proof.
- **D2** — named as the residual, inside the freeze's own recorded risk acceptance that *known-divergence-free is weaker than differentially clean*.

**Item 9 = LAPSED.** The frozen text says the `EnableTypeChecking` flip *"lands in the W2/W3 window"*. That window closed. The first adjudication re-read it as "the flip has a slot" and marked it ✅ — a reinterpretation in the favourable direction.

Cost **measured rather than guessed** — flipping the default and running `Calor.Compiler.Tests`: **92 failures / 6,158**, and they are type-checker *completeness* gaps, not stale fixtures:

| Diagnostic | Count | What |
|---|---:|---|
| `Calor0200` unknown type | 36 | `object` ×31, `Type` ×6 |
| `Calor0202` field access on non-record | 8 | member-access chains |
| `Calor0250`/`0251` | 4 | bind inference |

So the compiler would begin **rejecting programs that are valid today**. Whether that blocks v0.12.0 is a maintainer decision — not one an adjudication may make by reinterpretation. A further deferral must name a **new window**; the plan's register now carries the row so it lapses visibly rather than accruing silently.

## Three defects fixed — all on the shipped surface, not the one the evidence cited

**1. `CompileCalor` trusted a `.g.cs`'s existence, not its content** (item 4). `CompilationDriver` has always required the output's hash to match what the producing compile wrote, so a truncated or hand-edited generated file is a miss. The **MSBuild task real projects build through** only checked the file existed — and reported stale bytes as "up-to-date". Ported, with a regression test that corrupts the output plus a control proving an untouched output still skips. `EffectSummary != null` ported alongside as defence in depth (**not reachable today**: a null AST implies `HasErrors`, which returns before caching — stated rather than dressed up as a bug fix).

**2. `format --write`'s refusal ignored `--format json`** (item 7). Every other exit from `format` emits the envelope; the policy refusal wrote a bare stderr line with **empty stdout** and exited 1. An agent that asked for JSON got a parse failure instead of a policy decision — the one message whose entire purpose is to be understood was the one it could not read. Same class as the early-exit defects already pinned in `EnvelopeEarlyExitTests`, so the fix lives there.

**3. The containment had no test at all** (item 7). Nothing in the repo referenced `Calor1346`; item 7 was being adjudicated on code-reading alone. Three tests added — JSON refusal, text refusal, and **a control** proving the gate is an *acknowledgement* rather than a broken command, by running the write path through with `CALOR_EXPERIMENTAL_FORMAT_WRITE=1`. That needed an env-var overload on `CliTestHarness.RunCli`, scoped to the child process so it cannot leak across xUnit's parallel collections.

## Docs

`docs/cli/format.md` — the earlier audit fixed **3** sites documenting `--write` as an ordinary option. There were **12**, including the Quick Start block (the first thing a reader copies), the CI recipe, the editor-integration snippet, and the heal-in-place example. All corrected, plus a dedicated section covering the refusal, both escape hatches, the exit code, the JSON behaviour, and the sibling `CALOR_LSP_EXPERIMENTAL` containment.

## Filed, not fixed

`ContractTranslator`'s `StringInfo(bool IsNullable)` / `_stringInfo` is **written at five sites, read at none**, and `isNullable` is never once passed `true`. Inert — and it reads as a nullability guard that isn't one, which is worse than its absence. Recorded in the plan's deferred register.

## Verification

Solution-level `dotnet test -c Release --no-build`: **11 assemblies, ~8,100 passed, 0 failed**.

**Correcting a claim in this PR's first description.** It said solution-level `dotnet test` "only reports two" projects, so each had to be invoked by name, and quoted 8,485 across 12. The methodology claim is **false** — a solution-level run does pick up 11 assemblies. It came from a mis-scoped grep of my own output that I then repeated as a fact about the repo. The suite is green either way; the number and the reason given for it were not accurate.

Registered at **A-1.8**. Item 1's "published" half is dispositioned rather than asserted — CI evidences the *packaged* SDK's consumability from a local feed; "published" clears at the v0.12.0 publish event itself. **PP-W5 remains NOT adjudicated and may not be self-cleared.**

---

## Update — #872's first fix was itself incomplete

A second adversarial review found the D4 vector **still live** on the no-mode spelling, and it reproduced. .NET resolves `String.StartsWith(String)`, `EndsWith(String)` and `IndexOf(String)` to **CurrentCulture** (`Contains`/`Equals` are ordinal) while the solver models all of them ordinally — so omitting the comparison mode carried the identical defect on the form nearly everyone writes. Same signature: threw under `calor run`, printed the value under `--verify`, `calor verify` reporting 100% Proven. Also closed in #872.

This is recorded here because it **strengthens this PR's own thesis rather than undercutting it**: the first fix was narrow because it was scoped to *the divergence row as written* instead of *the property the row exists to protect*, and only the second is item 6's bar.

The verdict now carries the caveat it should have had from the start: **two rounds of review found two live vectors behind one green mark**, so "known-divergence-free" is a claim about how hard anyone looked — which is exactly what the freeze's own §2 risk acceptance says ("weaker than differentially clean").

### DIFF
--- a/docs/cli/format.md
+++ b/docs/cli/format.md
@@ -28,8 +28,8 @@ The `format` command formats Calor source files according to the canonical Calor
 # Format a single file (output to stdout)
 calor format MyModule.calr
 
-# Format and overwrite the file
-calor format MyModule.calr --write
+# Format and overwrite the file — GATED, see "Write mode is experimental" below
+calor format MyModule.calr --write --experimental
 
 # Check if files are formatted (for CI)
 calor format src/*.calr --check
@@ -53,7 +53,8 @@ calor format MyModule.calr --diff
 | Option | Short | Default | Description |
 |:-------|:------|:--------|:------------|
 | `--check` | `-c` | `false` | Check if files are formatted without modifying (exit 1 if not) |
-| `--write` | `-w` | `false` | Write formatted output back to the file(s) |
+| `--write` | `-w` | `false` | Write formatted output back to the file(s). **Refused (`Calor1346`) unless acknowledged** — see [Write mode is experimental](#write-mode-is-experimental) |
+| `--experimental` | — | `false` | Acknowledge that the `--write` path is experimental. Equivalent to `CALOR_EXPERIMENTAL_FORMAT_WRITE=1` |
 | `--diff` | `-d` | `false` | Show diff of formatting changes |
 | `--verbose` | `-v` | `false` | Enable verbose output |
 | `--heal` | — | `false` | Best-effort source-level repair of files too broken for the AST formatter. **Not semantics-preserving** — see [Heal Mode](#heal-mode) |
@@ -113,17 +114,42 @@ This allows you to preview changes before applying them.
 
 ---
 
-## Write Mode
+## Write mode is experimental
 
-Use `--write` to format files in place:
+`--write` is **disabled by the release policy** (#793/#760) and refuses with `Calor1346` unless you
+acknowledge it. The refusal is deliberate, not a bug: the formatter write path **can rewrite
+identifiers and drops comments**, so it is not safe to point at a codebase you care about.
+
+Read-only modes are unaffected and are what most workflows should use: `--check`, `--diff`, and the
+default stdout preview.
+
+To acknowledge, either pass the flag or set the environment variable:
 
 ```bash
 # Format single file
-calor format MyModule.calr --write
+calor format MyModule.calr --write --experimental
 
 # Format multiple files
+calor format src/*.calr --write --experimental
+
+# Or acknowledge once for a whole session / CI job
+export CALOR_EXPERIMENTAL_FORMAT_WRITE=1
 calor format src/*.calr --write
 ```
+
+Without the acknowledgement the command exits **1**, writes nothing, and reports:
+
+```
+error Calor1346: 'format --write' is disabled by the release policy (#793/#760): the formatter
+write path can rewrite identifiers and drops comments. Pass --experimental (or set
+CALOR_EXPERIMENTAL_FORMAT_WRITE=1) to acknowledge, or use --check/--diff for read-only formatting.
+```
+
+Under `--format json` the same refusal arrives as a normal envelope document carrying `Calor1346`,
+so an agent parsing stdout gets a policy decision rather than a parse error.
+
+The LSP has a sibling containment: the **formatting** and **rename** handlers register only under
+`CALOR_LSP_EXPERIMENTAL=1`. Every read-only LSP handler is unaffected.
 
 ---
 
@@ -159,8 +185,8 @@ reject:
 # Preview the healed source on stdout
 calor format Broken.calr --heal
 
-# Repair in place
-calor format Broken.calr --heal --write
+# Repair in place (--write is gated; see "Write mode is experimental")
+calor format Broken.calr --heal --write --experimental
 
 # CI / agent loop: report only (prints an ambiguousDecisions count per file)
 calor format Broken.calr --heal --check
@@ -280,7 +306,7 @@ The Calor formatter applies these rules:
 When formatting multiple files, errors in one file don't stop processing of others:
 
 ```bash
-calor format src/*.calr --write --verbose
+calor format src/*.calr --write --experimental --verbose
 ```
 
 Output:
@@ -302,7 +328,7 @@ Summary: 4 formatted, 1 error
 Use `--verbose` to see detailed processing information:
 
 ```bash
-calor format MyModule.calr --write --verbose
+calor format MyModule.calr --write --experimental --verbose
 ```
 
 Output:
@@ -331,10 +357,10 @@ Formatting MyModule.calr...
 
 ```bash
 # Find and format all .calr files
-find . -name "*.calr" -exec calor format {} --write \;
+find . -name "*.calr" -exec calor format {} --write --experimental \;
 
 # Or use shell globbing
-calor format **/*.calr --write
+calor format **/*.calr --write --experimental
 ```
 
 ### Pre-Commit Hook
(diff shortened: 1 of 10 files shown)


juanmicrosoft:
## Verdict changed: **PP-A1 = FAIL**

A third adversarial review round refuted this document's own D3 disposition, and it was right. I reproduced it before acting.

**D3 is a live false-`Proven`-elides vector.** The previous revision argued it unreachable because "Calor's `str` is non-nullable by construction". Every leg of that argument fails:

- **`str` is not non-nullable in practice** — `§B{bad:str}` binds `Environment.GetEnvironmentVariable`'s `null` with **zero diagnostics**.
- **No precondition is needed.** The divergence bites in the direction I never considered: Z3 makes `len(s)=0 ⟺ s=""` a tautology, while in C# `null` satisfies `IsNullOrEmpty` and `null == ""` is **false**.

```calor
§S (|| (! (isempty result)) (== result STR:""))    over    §R s
```

| | |
|---|---|
| `calor run` | **throws** |
| `calor run --verify` | prints `survived` — **check deleted** |
| `calor verify` | `Proven Rate: 100.0%` |

And it is not one form — `§S (>= (len result) INT:0)` crashes with a `NullReferenceException` unverified and survives verified. **Unlike D4 and D9 this is not closable by refusing an operation**: Z3's string sort has no null, so every total axiom of its string theory is unsound once a `str` holds one.

Two candidate fixes are recorded and **left to the maintainer**, because both are product decisions: enforce non-nullable `str` at the binder (correct, needs interop nullability analysis), or demote string-involving proofs to `Assumed` which never elides (D8 precedent — cheap, closes the whole class including undiscovered vectors, costs elision on every string postcondition).

A **fourth** vector was then found on #872 and is recorded there as **D12**: Z3 models strings as UTF-8 *bytes* while .NET counts UTF-16 *code units*, so `§S (== (len result) INT:2)` over `§R STR:"é"` is `Proven` and elided while false at runtime. It reproduces even when `:ordinal` is stated explicitly — i.e. following the D4 fix's own remediation advice.

### What this means

**v0.12.0 is blocked.** Three review rounds found three live vectors behind item 6's green mark, two of them behind rows I had personally re-audited. The registered recommendation: *known*-divergence-free is a claim about how hard anyone looked, and item 6's bar should be replaced by #779's solver-vs-runtime differential suite rather than re-asserted by another reading of the table. §2 of the freeze already says as much — known-divergence-free is **weaker than differentially clean**.

Everything else from the review is fixed: `lint --fix`'s refusal ignoring `--format` (the exact defect this PR fixed for `format --write`, at the sibling site), the refusal tests inheriting ambient env, the item-7 control that never proved a write, item 9's mis-tallied cost table (`char` 37 is dominant, not `object` 31; rows now sum to 92; residue includes checker *defects*), item 2's dropped third leg, item 6's dropped second clause, and item 3's overstated citation.

juanmicrosoft:
## Verdict updated: **items 1–8 PASS**, item 9 LAPSED

#876 merged, closing **D3** and **D12** by demoting string-carried proofs to `Assumed` (which never elides) on **both** elision channels — postconditions and refinement obligations. D3 was item 6's blocker, so item 6 now passes.

Rebased onto `main` and re-adjudicated.

### The caveat that belongs on this PASS

Four adversarial review rounds found four live false-`Proven`-elides vectors behind this one item — D4's explicit-mode half, D4's no-mode half, D3, and D12 — each time **after** it had been marked green, and two of them behind rows I had personally re-audited and argued safe.

**What finally closed it was not a better enumeration of the divergence table.** It was a change of mechanism: #876 removes *elision* for string-carried proofs, so the class is closed **by construction** rather than by having found every member. The trigger asks the translator whether a string sort was ever minted — the first attempt asked the contract AST and missed the function body, which review caught.

That is the honest reading of the bar. *Known*-divergence-free is a claim about how hard anyone looked, and on this evidence inspection does not find the bottom of the solver's string model. §2 of the freeze says as much in its own words: known-divergence-free is **weaker than differentially clean**.

**Registered recommendation:** item 6's bar should not be carried into the successor plan as "audit the table again". Replace it with #779's solver-vs-runtime differential suite, or with more closed-by-construction changes of the #876 kind.

### Still open, deliberately

- **D1, D2, D3, D12** remain open *modeling* gaps; only their elide consequences are neutralized. **#875** tracks D3's root cause (non-nullable `str` at the binder), which is what would let the demotion be lifted.
- **Item 9 = LAPSED** and referred to you. Measured cost of the flip: 92 failures / 6,160, dominated by type-checker completeness gaps (`char` ×37) plus real defects — valid programs would start being rejected.
- **PP-W5** is the other release gate, needs spend authorization, and may not be self-cleared.

PR #880, opened 2026-08-06
PP-W5 = PASS on w5-parity-002 (w5-parity-001 VOID) — the last v0.12.0 release gate

juanmicrosoft:
Adjudicates **PP-W5**, the last unadjudicated v0.12.0 release gate, on epoch **`w5-parity-002`**. The first attempt, `w5-parity-001`, is recorded **VOID**.

## Result: PASS

Control `v0.10.0` (`e24a6832`) vs treatment `main` (`87a783dd`), 4 registered N1 neutral pairs, 5 runs/arm, `raw` both arms, **interleaved**. 40/40 valid, 0 censored, 20/arm. **$54.08** ($92.83 cumulative across both attempts).

| pair | ratio (frozen metric) | ratio (corrected) |
|---|---:|---:|
| N1-001-string-utils | 0.470 | **0.953** |
| N1-002-inventory | 1.146 | 0.938 |
| N1-003-csv-row | 1.622 | **1.622** |
| N1-005-order-pipeline | 1.050 | 1.051 |
| **median point** | **1.0984** | **1.0016** |
| lower bound | 0.6531 | ≈0.81 |

PP-W5 fails only if **both** the lower bound exceeds 1.0 **and** the point exceeds 1.25. Neither fires on either basis → **PASS**, robust across 200 bootstrap seeds.

Iterations-to-green (row-mandated, omitted from the first record): control `{1:17, 2:2, 3:1}`, treatment `{1:18, 2:1, 4:1}`.

## Why attempt 1 was void

`--calor-dll` pins only the CLI — **which the agent never invokes.** It is told to run `dotnet build`, which binds Calor.Tasks/Sdk/Runtime and the emitter through the arm template's `__REPO_ROOT__`, and `run-m5-epoch.sh` never passed `--arm-repo-root`. Both arms compiled through the same `Calor.Tasks`. $38.75 for a non-measurement, and the CLI content-hash guard certified a contrast that did not exist.

**Attempt 2's contrast is verified by execution, not by flag:** `obj/calor/.calor-build-state.json`, written by `Calor.Tasks` itself, records `compilerHash` `65ee7cf0…` (control) vs `40472f37…` (treatment).

## A metric erratum, and a finding I withdrew

The frozen row pins output tokens to `agent.json` `usage.output_tokens`. **That field silently under-counts** on subagent-delegating and compaction-resumed runs — four runs across the two epochs, the worst by **55×** (543 recorded vs 30,084 actual).

Two claims from an earlier revision of this PR are **withdrawn**:

- *"one task cost the treatment half the tokens"* — the 0.470 does not exist; that run was the **most** expensive in its cell.
- *"a genuine contrast multiplied the between-pair spread sevenfold"* — corrected identically, the two epochs' spreads are 0.538 and 0.684 (**1.27×**), and an arm-label permutation test places the observed spread at the **50th percentile** of the no-effect null. No heterogeneity is claimed or supported.

Filed as **#881** (§7 erratum). The adjudicator transcribed the frozen metric faithfully; the metric is the defect.

## What the PASS means

**No large tax detected.** Corrected point 1.0016 — as close to parity as this design resolves — with a bound that still admits a wide interval. Not proof of equality: within-cell CV rose from 0.20 to 0.42 median, so realised power is **below** the frozen 0.33/0.62/0.87.

**No mechanism is attributed.** All four N1 fixtures contain **zero contracts**, so the D3/D12/D14 elision withdrawal cannot operate here; the `EnableTypeChecking` flip has zero supporting instances in the archived journals.

## Six apparatus defects

Five were **silent** — data loss, or a guard verifying the wrong artifact: arms overwriting each other; the M5 swap guard rejecting a valid parity config; the `raw` arm never invoking the agent on bash 3.2; the product never bound per arm (the void); interleaving overwriting run 1 with run 2; and the token under-count. Four were caught pre-spend or at $0.00.

Guards now hard-fail a parity epoch lacking per-arm roots with distinct `Calor.Tasks` hashes; every run stamps its root; the adjudicator refuses unless the pins record two distinct roots and every run's provenance matches one; and the §2 censoring caps it had dropped are restored.

## Consequence

Registered at **A-1.9**. Both v0.12.0 release gates stand: **PP-A1 = PASS** (A-1.8, #873), **PP-W5 = PASS** (A-1.9).


juanmicrosoft:
## The first epoch was VOID. Verdict re-run and restated.

Adversarial review found that `w5-parity-001` **measured nothing**, and I confirmed it independently before acting.

`--calor-dll` pins only the CLI — which the agent **never invokes**. It is told to run `dotnet build`, which binds Calor.Tasks/Sdk/Runtime and the emitter through the arm template's `__REPO_ROOT__`, and `run-m5-epoch.sh` never passed `--arm-repo-root`. **Both arms compiled through the same Calor.Tasks.** $38.75 for a non-measurement.

The proof is filesystem, not inference: neither arm worktree contained `src/Calor.Tasks/bin` at all, so had `__REPO_ROOT__` pointed there, every control run would have failed to build. All 20 were green.

Two of my own claims were independently wrong: the "arms are keyed on the build each run used" assurance (`calorDll` is the flag echoed back — it split a clean 20/20 and certified nothing), and the attribution paragraph invoking the D3/D12/D14 elision withdrawal, which **cannot operate here — all four N1 fixtures contain zero contracts**.

## `w5-parity-002`

Product genuinely split (`Calor.Tasks` `715cb5ad…` vs `9870805b…`, verified **before** the run), arms **interleaved**. 40/40 valid, 0 censored, 20/arm, **$54.08**.

| pair | control | treatment | ratio |
|---|---:|---:|---:|
| N1-001 | 12,238 | 5,752 | **0.470** |
| N1-002 | 8,403 | 9,633 | 1.146 |
| N1-003 | 14,466 | 23,468 | **1.622** |
| N1-005 | 7,213 | 7,577 | 1.050 |

**Point 1.0984, lower bound 0.6531** → neither gate fires → **PASS**.

## The finding that outweighs the verdict

Per-pair ratios span **0.470 to 1.622**, no consistent direction. The void epoch is an accidental clean null (same compiler both arms): ratios 0.987–1.159, spread 0.17. **A genuine contrast multiplied the between-pair spread ~7×.**

So PASS means only *no large **systematic** tax detected*. The bound admits anything from a ~35% improvement to a ~10% tax. **This epoch does not establish parity and could not have** — and the row's power figures were calibrated on the null variance at `m5-compare-001`, so against realised variance here, true power is **lower** than frozen. No mechanism is attributed.

## Five apparatus defects, four silent

Arms overwriting each other; the M5 swap guard rejecting a valid parity config; the `raw` arm never invoking the agent on bash 3.2; **the product never bound per arm** (the void); and interleaving re-invoking `run-pair` so run 2 overwrote run 1. Four caught pre-spend or at $0.00.

Guards now **hard-fail** a parity epoch lacking per-arm roots, and the adjudicator refuses unless the pins record two distinct roots with distinct `Calor.Tasks` hashes and every run's provenance matches one.

Also added: iterations-to-green (row-mandated, omitted from the void record), §2 censoring caps (silently dropped by the adjudicator), and realised-vs-projected spend with the reason for the gap.

Issue #762, opened 2026-07-24
P0: Make parser, AST, binder, and overload resolution structurally total


juanmicrosoft:
## Summary

Make parsing, AST dispatch, binding, and overload resolution structurally total so every accepted expression retains its children, type, and symbol identity for later analyses.

## Evidence and impact

- `Binder.BindExpression` handles only a minority of concrete expression nodes.
- Unsupported expressions become a zero-child `<unsupported:...>` call, hiding nested calls, division, indexing, overflow, and taint.
- Casts return the original incorrectly typed operand.
- `is` and pattern tests become literal `true`.
- Decimals are downcast to `double`.
- Top-level function registration ignores failed `TryDeclare` results and does not use the overload-set machinery used for class members.
- Parser expression-start classification and expression dispatch disagree.
- Several concrete `AstNode` subclasses implement no-op/default `Accept`.

These gaps make downstream verification and diagnostics unsound even when code generation happens to work.

## Required implementation

1. Add structural bound nodes for every accepted expression family, including conversion, type/pattern tests, arrays/indexes, collections, lambdas, await, coalesce, match, quantifiers, string operations, and interop.
2. Preserve every child expression and an explicit stable type.
3. If a node remains unsupported for analysis, retain its children and emit an `analysis incomplete` diagnostic; never silently erase the subtree.
4. Preserve decimal precision/type and conversion target types.
5. Register top-level functions in overload sets and resolve using full signatures.
6. Diagnose duplicate signatures, ambiguity, and no matching overload.
7. Derive expression-start classification and parser dispatch from one authoritative table.
8. Give every concrete visitable AST node real visitor dispatch, or reclassify helper/value objects so they are not `AstNode`.
9. Add stable `SymbolId` identities that can be consumed by analyses and LSP.

## Required regression coverage

- Reflection-generated test covering every concrete `ExpressionNode` and `AstNode`.
- A diagnostic seed nested inside every expression wrapper and proof that it remains reachable.
- Cast, type-test, pattern-test, decimal, index, and collection binding.
- Overloads by arity and type; duplicates, ambiguity, and no-match cases.
- Every expression-start token in return, bind, argument, and nested contexts.
- Unsupported-node diagnostics that name the exact node and retain child traversal.

## Success metrics / definition of done

- 100% of accepted expression kinds structurally bind.
- No accepted program silently produces a zero-child `<unsupported:...>` expression.
- Every bound expression has a stable type and retains all semantically evaluated children.
[...]

PR #892, opened 2026-08-10
fix(parser): keyword args on generic calls degrade to a diagnostic, not an NRE (#874)

### LINKED ISSUE #874: ContractTranslator: NRE when a call expression carries a keyword argument inside a contract
Found during adversarial review of #872. **Pre-existing — reproduces identically on `main`, not introduced by that PR.**

A *call expression* (capitalized, so it parses as `CallExpressionNode` rather than `StringOperationNode`) carrying a keyword argument inside a `§Q`/`§S` crashes the compiler instead of producing a diagnostic:

```calor
§Q (StartsWith s STR:"hello" :ignore-case)
```

```
Error processing pre.calr: Object reference not set to an instance of an object.
```

The report comes back empty. Expected: a normal diagnostic (an unsupported-form `Calor0718`, or a parse/bind error naming the keyword argument) — never an unhandled `NullReferenceException` with no location.

Two things make this worth fixing beyond the crash itself:

1. **It is reachable from a plausible mistake.** The lowercase `starts`/`contains` forms are the documented spelling, but the capitalized names are the C# ones, and someone porting a contract from C# will write them. They get a crash rather than a "not supported here" message.
2. **A crash is the one outcome that carries no information.** Every other unsupported form in the verifier degrades to `Unsupported` with the runtime check retained. This path degrades to nothing.

Related context: #872 closes divergence D4 by refusing non-ordinal and no-mode string comparison in contracts. This issue is adjacent — same surface, different failure mode — but independent of it.

juanmicrosoft:
## Summary

v0.13 quick-wins batch, item 1 (roadmap §2.1 truth floor). Fixes #874 — a call expression carrying a keyword argument inside a `§Q`/`§S` crashed the compiler with an unlocatable `NullReferenceException` and an empty report.

**The issue's own diagnosis was wrong, and the investigation corrected it:** the ContractTranslator was never the crash site — it degrades `CallExpressionNode` to `null` (Unsupported) by design. The real chain:

1. `(StartsWith s STR:"hello" :ignore-case)` — the capitalized C# spelling doesn't match the lowercase string-op table, so the parser's PascalCase recovery path builds a generic `CallExpressionNode` **with the `KeywordArgNode` still in its argument list**, violating the node's documented "internal to parsing, never escapes" invariant.
2. `ExpressionSimplifier` rebuilds contract argument lists via `arg.Accept(this)`; `KeywordArgNode.Accept<T>` returns `default!` (null) — injecting a **null argument** into the rebuilt call.
3. `CSharpEmitter.Visit(CallExpressionNode)` dereferences the null argument while emitting the precondition's runtime guard (`CSharpEmitter.cs:3107`).

## Fix (two layers)

- **Parser — enforce the invariant where it lives**: all three generic-call recovery sites route through `RejectKeywordArgs`, which reports `InvalidLispExpression` naming the keyword (`Keyword argument ':ignore-case' is not supported on call 'StartsWith'…`) and pointing at the lowercase forms — the plausible-mistake path (porting a contract from C#) now gets an actionable message. Protects **every** downstream consumer (simplifier, emitter, translator, IdScanner) at once.
- **ExpressionSimplifier — defense in depth**: all three identical argument-rebuild loops now preserve the original node when `Accept` returns null. This matters beyond `KeywordArgNode`: **five** AST classes have no-op `default!` Accepts (`KeywordArgNode`, `ControlFlowNodes.cs:156`, `TypeNodes.cs:75/127/147`) — the #762 "no-op Accept" evidence bullet quantified; the other four are the same latent null-injection hazard, to be retired structurally by the #762 rebuild under F-1's item-8 decision.

## Verification

- Two pinning tests (`KeywordArgContractTests`, `§Q` and `§S` positions); **both fail pre-fix** with the exact NRE from the issue (discriminating pin), and the `§Q` test asserts the diagnostic names the keyword.
- CLI repro from the issue now prints the located diagnostic instead of `Error processing pre.calr: Object reference not set…` with an empty report.
- Full suite: **8,273 tests, 0 failed, 17 skipped** (the known LSP/MCP skip profile) — the lowercase string-op keyword path (`:ordinal`, `:ignore-case` on `starts`/`contains`/…) is untouched and its existing tests stay green.

Fixes #874

### DIFF
--- a/src/Calor.Compiler/Ast/StringOperationNode.cs
+++ b/src/Calor.Compiler/Ast/StringOperationNode.cs
@@ -84,11 +84,12 @@ public sealed class KeywordArgNode : ExpressionNode
     }
 
     // KeywordArgNode is internal to parsing - it should be consumed by the parser
-    // and not appear in the final AST. The parser enforces this: string operations
-    // consume comparison-mode keywords, and the generic call recovery paths reject
-    // any that remain with a diagnostic (Parser.RejectKeywordArgs, #874). These
-    // methods should therefore never be called; default! here means any escape
-    // becomes a null child in a visitor rebuild, so the invariant must hold.
+    // and not appear in the final AST. The parser enforces this at a single choke
+    // point (Parser.FilterKeywordArgs, #874): a trailing keyword on a recognized
+    // string operation is consumed as the comparison mode; every other occurrence
+    // is rejected with a diagnostic and dropped. These methods should therefore
+    // never be called; default! here means any escape becomes a null child in a
+    // visitor rebuild, so the choke point is load-bearing.
     public override void Accept(IAstVisitor visitor) { }
     public override T Accept<T>(IAstVisitor<T> visitor) => default!;
 }
--- a/src/Calor.Compiler/Ast/StringOperationNode.cs
+++ b/src/Calor.Compiler/Ast/StringOperationNode.cs
@@ -84,7 +84,11 @@ public sealed class KeywordArgNode : ExpressionNode
     }
 
     // KeywordArgNode is internal to parsing - it should be consumed by the parser
-    // and not appear in the final AST. These methods should never be called.
+    // and not appear in the final AST. The parser enforces this: string operations
+    // consume comparison-mode keywords, and the generic call recovery paths reject
+    // any that remain with a diagnostic (Parser.RejectKeywordArgs, #874). These
+    // methods should therefore never be called; default! here means any escape
+    // becomes a null child in a visitor rebuild, so the invariant must hold.
     public override void Accept(IAstVisitor visitor) { }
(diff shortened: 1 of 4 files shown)


juanmicrosoft:
Adversarial review round applied. The review (verdict 90%, NEEDS-FIXES) **demonstrated** five still-crashing routes past the original three-site fix plus a silent-invalid-codegen route — the fix now enforces at the production site instead: a `FilterKeywordArgs` choke point at operator-argument completion, and `ParseLispArgument` minting `KeywordArgNode` only for the opted-in main loop (closing the implication/quantifier/cast side entrances). Simplifier `??` guards reverted per review (palliative, inconsistent coverage). Tests 2 gated → 9 ungated, covering every demonstrated route; all six review repro files produce named diagnostics, zero crashes; suite 8,280/0.

Issue #789, opened 2026-07-24
P1: Make builds and native dependencies hermetic and supply-chain verified


juanmicrosoft:
## Summary

Make Calor builds and releases hermetic and supply-chain verifiable. Stop downloading unauthenticated native binaries during normal builds and stop mutating checked-in source resources as a build side effect.

## Evidence and impact

- `Calor.Compiler.csproj` automatically downloads Z3 when local files are absent.
- Download scripts and release workflow use `curl` without committed checksum/signature verification.
- The release workflow downloads project-release binaries and packages them without provenance validation.
- A deps.json post-processing target hard-codes `.NETCoreApp,Version=v8.0` while the project targets .NET 10 and ignores failures.
- The normal build copies canonical E2E files into checked-in `src/Calor.Compiler/Resources/SelfTest`.
- Dependency versions are decentralized and lock files are absent.

These behaviors make builds network-dependent, nonhermetic, vulnerable to poisoned artifacts, and capable of dirtying the source tree.

## Required implementation

1. Produce or consume Z3 artifacts with committed SHA-256/SHA-512 checksums and verifiable provenance/signatures.
2. Make all download commands fail on HTTP errors and verify size/hash before use.
3. Remove network downloads from ordinary build targets; restore dependencies in an explicit bootstrap/restore phase.
4. Prefer proper NuGet/native runtime asset packaging over post-hoc deps.json mutation.
5. Remove the hard-coded target framework key and add automated deps/runtime resolution tests.
6. Generate self-test resources under `obj/` or enforce synchronization in a separate explicit command; normal build must not write tracked source.
7. Add SBOM/provenance artifacts to releases.
8. Centralize dependency versions and evaluate lock-file/locked restore for release builds.
9. Test offline rebuilds from a fully restored dependency cache.

## Required regression coverage

- Corrupt hash, truncated download, HTTP 404/error page, and unexpected asset size.
- Offline build after explicit restore.
- Build with no preexisting Z3 files.
- Runtime load tests on all supported RIDs.
- Assert `git status` is unchanged after restore/build/test/pack.
- Dependency-lock mismatch and release provenance validation.

## Success metrics / definition of done

- Normal `dotnet build --no-restore` performs zero network access.
- Every shipped native binary has a verified committed checksum and provenance record.
- A corrupt or missing asset fails before compilation/packaging.
- Build/test/pack leave the tracked worktree byte-for-byte unchanged.
- No build logic contains a hard-coded obsolete TFM key.
[...]

PR #893, opened 2026-08-10
fix(emitter): maintain the elision cursor in Visit(MethodNode) (#879)

### LINKED ISSUE #879: Elision is keyed by a mutable cursor that Visit(MethodNode) never maintains — class-method postconditions never elide, and §MT violations report the wrong function id
Found by adversarial review of #878. **Latent for soundness, live for two other defects.** Filed as follow-up rather than fixed there, on the reviewer's recommendation.

## The mechanism

`CSharpEmitter` elides a postcondition via `GetPostconditionVerificationResult()`, keyed on `(_currentFunctionId, _currentPostconditionIndex)`. Those two fields are assigned in exactly **two** places — `Visit(FunctionNode)` and `EmitExtensionMethod`.

**`Visit(MethodNode)` sets neither**, yet it calls `Visit(ensures)`.

So the elision key is a *mutable cursor*, and there is an emission path that does not maintain it.

## What that costs today

Identical contract, identical body, two forms, same file:

```
Ok.Id  (mt001, §MT)   Postcondition 0: Proven  →  if (!(__result__ >= 0)) throw … "unknown"
Same   (f001, §F)     Postcondition 0: Proven  →  // PROVEN: Postcondition statically verified
```

1. **Class-method postconditions never elide, even when `Proven`.** An undocumented capability gap on exactly the OO shape real code uses — and it means the "what still proves" measurements taken during the D14 work reflect no methods at all.
2. **Every `§MT` contract violation reports the wrong function id** — `"unknown"` normally, and a *foreign* id when an enum extension precedes the class (reproduced: a `Bad.Wrong` postcondition exception carried `"f001"`, the `§EEXT` function's id). That misattributes a production failure to an unrelated function.

## Why it is not live as a soundness hole — and how thin that is

If `_currentFunctionId` ever points at a **verified** function while a method's `§S` is emitted, that method's check is deleted on an unrelated function's proof. It doesn't happen today, for **two independent accidents**:

- Within a module the emission order is interfaces → enums → enumExtensions → delegates → classes → interop → typePP → functions, so the only id-setter that can precede a class is `EmitExtensionMethod`; and
- `ContractVerificationPass.EnumerateContractBearers` covers only `module.Functions` and class methods — **never `EnumExtensions`** — so that lookup returns null.

**Verifying `§EEXT` contracts — a natural, small change — makes it live.**

## The class this belongs to

Worth naming, because it is *not* the class the last five vectors were in. Those were all "a false `Proven` deletes a check". This one is **"the elision is keyed by a mutable cursor, and the key is not maintained on every emission path"** — a different failure mode reaching the same consequence.

## Fix

`_currentFunctionId = node.Id; _currentPostconditionIndex = 0;` at the top of `Visit(MethodNode)`.

**It must not land alone.** The fix *enables a new elision surface*, so it needs a test that a method postcondition on an array-taking method comes out `Assumed` — which is only safe now that D14 (#878) is in.

## Ride-along

`ProofOutcome.ToObligationStatus()` maps `Assumed → Timeout`, and `Timeout` happens to be one of three arms that emit a guard; `Unsupported`/`Pending` fall through to `// TODO` with **no** guard. `Assumed`'s guard survival is therefore a coincidence rather than a pinned invariant. Pre-existing and not worsened by #878, but it belongs with this work since it touches the same emitter region.

juanmicrosoft:
## Summary

v0.13 quick-wins batch, item 2 (roadmap §2.1 truth floor). Fixes #879: the postcondition-elision key is a mutable cursor that `Visit(MethodNode)` never maintained — class-method postconditions never elided even when `Proven`, and `§MT` contract violations reported `"unknown"` (or a foreign id) as the failing function. The near-miss soundness route (a method's check deleted on an unrelated function's proof, live the moment `§EEXT` contracts get verified) is closed by the same two lines.

## Changes

- `Visit(MethodNode)` sets `_currentFunctionId`/`_currentPostconditionIndex`, with a comment stating the invariant every contract-visiting emission path must maintain.
- **Ride-along from the issue**: `ObligationStatus.Unsupported` joins the guard-emitting arms — previously it fell through to a `// TODO` comment with **no guard**, and `Assumed` kept its guard only by coincidence of its `Timeout` mapping. An obligation the solver cannot model keeps its check (#779 posture).
- **Golden pinning the defect** (the D15 lesson recurring): `08_contract_inheritance_z3` asserted the `"unknown"` ids. Corrected at the E2E scenario source — the build's `SyncSelfTestFiles` target overwrites `Resources/SelfTest` from `tests/E2E/scenarios/`, which is why regenerating the resource copy alone silently reverted (a live demonstration of the #789 sync finding).

## Verification

- Four pinning tests (`MethodElisionCursorTests`): the two cursor pins **fail without the fix** (revert-verified); the two new-surface pins hold the opened elision surface to the demotion/refusal machinery.
- **Honest deviation from the issue's landing requirement**: it asked for "array-taking method comes out `Assumed`" — at the e2e surface that shape is *refused* at translation (`Calor0718`, guard kept) because lisp `(len x)` models string length only; the demotion family is instead pinned via the string-`:ordinal` form (`Calor0720` `Assumed` on `§MT` exactly as on `§F`, mirroring the IntegrationTests precedent).
- Golden diff verified line-by-line: exactly the four id corrections (`"unknown"` → `mt001`/`mt002` — the fix also corrects **pre**condition attribution) plus trailing-newline normalization.
- Full suite: **8,284 tests, 0 failed, 17 skipped** (known profile).

Fixes #879

### DIFF
--- a/src/Calor.Compiler/CodeGen/CSharpEmitter.cs
+++ b/src/Calor.Compiler/CodeGen/CSharpEmitter.cs
@@ -2738,6 +2738,16 @@ public sealed class CSharpEmitter : IAstVisitor<string>
 
     public string Visit(MethodNode node)
     {
+        // #879: the postcondition-elision key is a mutable cursor
+        // (_currentFunctionId, _currentPostconditionIndex) that every emission path
+        // visiting contracts MUST maintain. Before this was set here, class-method
+        // postconditions never elided even when Proven, §MT contract violations
+        // reported "unknown" (or a foreign id) as the failing function, and — had
+        // §EEXT contracts ever been verified — a method's check could have been
+        // deleted on an unrelated function's proof.
+        _currentFunctionId = node.Id;
+        _currentPostconditionIndex = 0;
+
         // Clear declared variables tracking for new method scope
         ResetDeclScopes(node.Parameters);
 
@@ -5712,6 +5722,12 @@ public sealed class CSharpEmitter : IAstVisitor<string>
                     case Verification.Obligations.ObligationStatus.Boundary:
                     case Verification.Obligations.ObligationStatus.Failed:
                     case Verification.Obligations.ObligationStatus.Timeout:
+                    // #879 ride-along: Unsupported keeps its runtime guard by design.
+                    // Previously it fell through to the no-guard TODO comment — and
+                    // Assumed's guard survived only because ToObligationStatus maps it
+                    // onto Timeout. An obligation the solver cannot model must keep its
+                    // check (#779 posture: guards stay until verification proves).
+                    case Verification.Obligations.ObligationStatus.Unsupported:
                         // Emit runtime guard
                         var condition = node.Condition.Accept(this);
                         AppendLine($"if (!({condition})) throw new InvalidOperationException(" +
--- a/src/Calor.Compiler/CodeGen/CSharpEmitter.cs
+++ b/src/Calor.Compiler/CodeGen/CSharpEmitter.cs
@@ -3322,6 +3322,14 @@ public sealed class CSharpEmitter : IAstVisitor<string>
 
     public string Visit(PropertyAccessorNode node)
     {
+        // #879: property accessors are never verified (EnumerateContractBearers covers
+        // functions and class methods only), and this node carries no id. A stale cursor
+        // here would attribute an accessor precondition failure to a foreign method — or,
+        // worse, elide on a foreign proof. Null is the honest key: lookup misses, guard
+        // kept, id reported as "unknown".
+        _currentFunctionId = null;
+        _currentPostconditionIndex = 0;
+
         var accessorKeyword = node.Kind switch
         {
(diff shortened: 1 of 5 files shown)


juanmicrosoft:
Adversarial review round applied — and this one mattered: the review (verdict 12%) **demonstrated that the first commit made #879's latent soundness hole live** — operators emit after methods, inherited the verified mt-id cursor, and a false, never-verified operator postcondition was reproduced eliding on a foreign method's proof. Constructors/accessors similarly upgraded "unknown" to confidently-wrong foreign ids. Fixed by maintaining the cursor on every unverified bearer (operator/ctor → own id, accessor → null), completing the ride-along (Pending joins Unsupported in the guard arms; both now pinned — the original hunk was revert-invisible), and strengthening the demotion pins with guard-presence assertions. Both new pins fail against the armed intermediate state (stash-verified). Suite 8,288/0.

PR #894, opened 2026-08-10
feat(verify): proof-based guard elision is now opt-in (roadmap v0.13 §2.1)

juanmicrosoft:
## Summary

v0.13 quick-wins batch, item 3 — the elision policy decision the roadmap takes by design rather than by gate (§2.1): **proof-based guard elision becomes explicitly opt-in; verification is diagnostic by default.**

- Default: a `Proven` postcondition or `Discharged` `§PROOF` obligation **keeps its runtime check**; the verdict is reported, never silently acted on.
- Opt-in: `--elide-proven-guards` (CLI, threaded through `CompileAsync`/`CompileCore`), `CompilationOptions.ElideProvenGuards`, MSBuild task property `ElideProvenGuards`.
- The `run`/`test` execution paths (`ExecutionWorkspace`) deliberately have no flag — execution always keeps guards.
- Both elide sites gated: the postcondition `Proven` branch and the obligation `Discharged` arm (`Discharged` without opt-in now joins the guard-emitting group alongside the #893 ride-along statuses).

## Why

Seven false-`Proven` elide vectors closed across v0.10–v0.12; D15 proved the demotion mechanism is structurally blind to emitter-side vectors; the #893 review demonstrated a *new* foreign-proof elision route two days after the class was thought understood. The differential gate that would justify default-on (freeze registration F-4: zero mismatches over the pinned `verification-modeled-forms.md` denominator, coverage fraction published) is registered but not yet built — roadmap §3.5 gate 6 is the re-enable bar.

## Verification

- Existing elision pins (IntegrationTests, OutcomeCorpusTests, MethodElisionCursorTests, ObligationTests) opt in explicitly — they pin the elision machinery, which is unchanged behind the flag.
- New default-off pins: `ProvenPostcondition_WithoutOptIn_KeepsGuard` and `CSharpEmit_DischargedObligation_DefaultKeepsGuard`.
- Full suite: **8,290 tests, 0 failed, 17 skipped** — nothing outside the explicitly opted-in pins depended on the old default, which is itself evidence the flip's blast radius is as narrow as intended.
- CHANGELOG Unreleased entry included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+- **Proof-based guard elision is now opt-in** (`--elide-proven-guards` on the CLI,
+  `ElideProvenGuards` on `CompilationOptions` and the MSBuild task). By default a `Proven`
+  postcondition or `Discharged` `§PROOF` obligation keeps its runtime check — verification
+  verdicts are diagnostic. This executes roadmap v0.13 §2.1: the elide surface has produced
+  seven false-`Proven` vectors to date and its differential gate (freeze registration F-4) is
+  not yet built; elision re-enables by default only when that gate is green (roadmap §3.5
+  gate 6). The `run`/`test` execution paths always keep guards regardless of the flag.
+
 ### Corrected
 - **The v0.12.1 notes below understated the VS Code outage by about six times, and implied a
   publish that did not happen.** They said the extension "can be published again" after "three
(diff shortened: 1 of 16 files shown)


juanmicrosoft:
Adversarial review round applied (verdict 85%). The flip's gate completeness survived a full sweep — both elide sites gated, mutation-checked (gate removed → both default-off pins fail), CLI verified e2e both ways, run/test always-guards claim confirmed true. Fixes applied: **Sdk.targets now actually passes `CalorElideProvenGuards`** (the MSBuild opt-in was advertised but unwired); four contradicting doc surfaces corrected with the **F-4 denominator re-pinned by additive amendment** (prose-only change, form content unchanged); the default-off pin asserts a genuine Proven verdict — which exposed one more accuracy defect the review missed: the verbose Calor0713 message claimed "Runtime check elided" unconditionally, now conditional on the flag; no-op flag combination warns; contract-mode-off obligation wrinkle in CHANGELOG; indentation fixed. One finding declined with reason (automated CLI-plumbing test — no harness infra to host it; manual e2e verified). Suite 8,290/0.

PR #895, opened 2026-08-10
feat(ci): PP-S4 fixture registry + d-s1.5-fixtures gate (A-1.5.7, roadmap gate 6)

juanmicrosoft:
## Summary

v0.13 quick-wins batch, final item — discharges the Call S debt (*"the first `SupportLevel` promotion re-arms PP-S4 with no CI to enforce it"*, `call-s-adjudication.md:100-103`) by building PP-S4's registered instrument at **all three frozen parts of A-1.5.7**: location, schema, and CI entry point.

- **Registry**: `bench/phase0-agent-native/fixtures/d-s1.5/` + schema README. Initial contents empty, per the registration ("frozen by location, schema, and entry point, not by content").
- **Detection** (`scripts/check-d-s1.5-fixtures.sh`): `SupportLevel` promotions toward `Full` — including a new entry born above `NotSupported`, the one-step form of the same maneuver — and **net** `ConversionLossKind` removals aggregated across the whole Migration diff (moving a `RecordLoss` between files nets zero; no false fire).
- **The discriminating pin runs continuously**: the CI job executes the script's three-pin `--self-test` before every real check — a synthetic fixture-less promotion must fail, a no-change diff must pass, an uncertified loss-kind removal must fail. Roadmap §2.5 gate 6 asks for the pin to be "demonstrated"; this executes it every run.
- **Greenness** (`DS15FixtureRegistryTests`): per-fixture structure, exact conversion match against `expected.calr`, and compilation of both Calor files — **indeterminate (won't build) fails rather than skips**, implementing the registration's conservative rule.

## Consequence

Roadmap §3.2's sequencing constraint — no `FeatureSupport` promotion or loss-ledger removal merges before this gate is green — is now mechanically enforceable, closing the silent-roll-forward risk the substrate plan flags (§7 item 4: "fidelity bought by relabelling corrupts the evidence base").

## Known evasion surfaces for review

Named so the reviewer verifies rather than rediscovers: key-rename-while-promoting (caught: rename = new key born at `Full`), two-PR delete-then-readd (each PR gated independently; the re-add is a birth above `NotSupported` → gated), force-push base-SHA on push events (falls back to `HEAD~1`), awk extractor's coupling to `FeatureSupport.cs` formatting.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,4 +5,6 @@
 **/*.cs @juanmicrosoft
 /scripts/check-calor-first-diff.sh @juanmicrosoft
 /scripts/check-calor-first-diff-tests.sh @juanmicrosoft
+/scripts/check-d-s1.5-fixtures.sh @juanmicrosoft
+/bench/phase0-agent-native/fixtures/d-s1.5/ @juanmicrosoft
 /.github/workflows/test.yml @juanmicrosoft
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -132,6 +132,54 @@ jobs:
             exit 1
           fi
 
+  d-s1.5-fixtures:
+    # PP-S4's instrument (A-1.5.7; roadmap v0.13 §2.5 gate 6): every SupportLevel
+    # promotion toward Full and every net ConversionLossKind removal must carry a
+    # registered fixture. The script's --self-test runs FIRST on every invocation,
+    # so the gate is demonstrated to fail on a fixture-less promotion continuously,
+    # not once (the discriminating pin, per pins-must-discriminate).
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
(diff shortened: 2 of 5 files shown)


juanmicrosoft:
Adversarial review round applied (verdict 15% — the harshest of the batch, deservedly). **CRITICAL-1 would have disabled all of CI on merge**: dots are illegal in job ids and the pushed workflow failed in 0s with zero jobs — fixed (`ds15-fixtures` id, `d-s1.5-fixtures` display name = the A-1.5.7 entry point), and workflow validation is now part of the loop. All four demonstrated evasions closed (non-literal Support rewrite → fail-closed set-diff + entry-count legs; space-bearing keys → tab-separated maps + sanitization rule; comment channels → stripped in both scanners with a negative-control pin; file-move false-fire → tree-wide per-ref counting). CODEOWNERS now covers the script + registry. Self-test grew to five pins. README carries the disclosed-limits section and the CRITICAL-2 note: the required-status-check addition is an admin action to perform at merge, on the record. Remaining accepted residuals are exactly the README's disclosed list.

PR #898, opened 2026-08-10
docs: #762 binder rebuild scoping — the v0.13 spine (Draft v1)

juanmicrosoft:
## Summary

Scoping doc for the #762 binder rebuild — the roadmap's load-bearing spine (§2.1 MUST) and the item v0.14/v0.15 cannot proceed without. Per the freeze discipline, this merges **before** any binder code: the F-2 measurement instrument lands in the first implementation PR (B1) so the gap is visible and measured before any of it closes.

Grounded in the code, not the issue text: 19 of 61 expression classes bind today; the decimal→double downcast is visible in the switch; the fallback's own comment records the false-positive class the replacement must preserve opacity against; and the bound tree's consumers are exactly the analysis layer — emitters are AST-based, so codegen is out of blast radius.

Key structural decisions: one authoritative dispatch table with a reflection completeness test (the F-1 bidirectional check's code-side half); `BoundIncompleteExpression` that **retains bound children** instead of erasing subtrees; SymbolIds from the already-validated syntax ids with exact identifier spans (the substrate for the roadmap's index/rename gates, and the proper retirement of the #879/#893 mutable-cursor class); types stay strings in 0.13 with the 0.14 typed-representation boundary explicitly fenced as a non-goal.

Eight-PR slicing with per-PR exit criteria, checker-delta disclosure per family (five checkers key on the fallback shape today — their verdict movement is measured against a B1 baseline snapshot, not discovered), and four named risks.

### DIFF
--- a/docs/plans/binder-rebuild-762-scoping.md
+++ b/docs/plans/binder-rebuild-762-scoping.md
@@ -1,33 +1,42 @@
 # #762 binder rebuild — scoping (v0.13 spine)
 
 **Date:** 2026-08-10
-**Status:** Draft v1 — for adversarial review before any binder code merges
+**Status:** Draft v2 — adversarial review round 1 applied (verdict 65%; both CRITICALs and all
+Majors/minors addressed; dispositions in §8)
 **Governing inputs:** issue #762 (DoD: 100% of accepted expression kinds structurally bind),
 `v0.13-freeze-registrations.md` F-1 (the 61-class tier assignment — 55 registered / 6 residual —
 frozen before this work; the denominator of roadmap §2.5 gate 1) and F-2 (the measurement corpus),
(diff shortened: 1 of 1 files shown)


juanmicrosoft:
Review round 1 applied (verdict 65%). The two CRITICALs materially changed the plan: the consumer map was wrong — **the LSP binds every open document with the live diagnostics bag** (D2's diagnostic would have hit every editor from B1) while **the analyze path discards its binder bag** (the instrument would have read nothing), so B1 now carries diagnostic routing + an Info-until-B8 severity decision, and §7 defines the exit criterion against where binding actually runs. SymbolIds are scheduled (B1) and corrected (file-path component — module ids demonstrably collide; declaration-order local keys — per-name ordinals renumber survivors under the rename gate's own operation). D2 got explicit two-phase semantics resolving the B1-exit contradiction. The frozen-marker change is handled by additive amendments in B1 with the classification argued. All counts corrected. Full disposition in §8; declined: none.

PR #900, opened 2026-08-10
feat(binding): make parser, AST, and binder structurally total (#762)

### LINKED ISSUE #762: P0: Make parser, AST, binder, and overload resolution structurally total

juanmicrosoft:
## Summary
- unify parser expression-start classification and dispatch behind one authoritative registry
- make every concrete visitable AST node dispatch exactly once and reclassify parser-only keyword arguments
- structurally bind all 60 accepted expression families with stable types and retained evaluated children
- preserve decimal, conversion, type/pattern, collection, initializer, control-value, unsafe, and interop semantics
- add stable source-qualified `SymbolId`s and exact overload resolution for top-level functions and members
- migrate dataflow, bug-pattern, taint, effect, contract, verification, call-graph, and LSP consumers to exhaustive traversal and symbol identity
- add reflection/completeness gates that fail when new AST/expression/bound-switch coverage is missing

## Behavioral guarantees
- no accepted expression silently becomes a zero-child `<unsupported:...>` call
- unsupported/incomplete analysis retains children and emits an exact `AnalysisUnsupportedNode` diagnostic
- duplicate signatures, ambiguous overloads, and no matching internal overloads are explicit diagnostics
- internal calls carry canonical resolved function symbols; external calls remain explicitly unresolved
- shadowed variables, overloads, and identical structural IDs in different files do not alias

## Validation
- `dotnet test --no-restore`: 8,506 passed, 17 skipped, 0 failed
- full solution builds performed throughout with 0 warnings/errors
- acceptance audit found no #762 blockers against current `main`

Closes #762


juanmicrosoft:
Closing as **superseded by the B1–B8 incremental rebuild** (#762, scoping doc `docs/plans/binder-rebuild-762-scoping.md`, final merge #911 / `55a4c920`). The B-series landed the same #762 scope in eight adversarially-reviewed slices, and gate 1 (§2.5 binding totality) is measured green on both F-2 legs.

This PR was far from wasted — B8 ported five of its designs with attribution, and they all survived adversarial review: the `ExpressionParsers` one-table unification, the `ParseAttributes(maxGroups: 1)` brace-initializer fix, the seven item-8 visitor implementations, the DispatchProxy + `GetUninitializedObject` architecture test, and the token-context matrix test.

The roughly half of this diff that addresses LATER roadmap phases — SymbolId machinery and consumers, full-signature overload resolution, the exhaustive checker-traversal rewrites, the resolved call graph, and the LSP cross-file/rename batteries — is inventoried in **#913** so it gets diff-read before any of that work is re-implemented from scratch (branch tip preserved at `0e99c4a9`).

PR #902, opened 2026-08-10
feat(binder): #762 B2 — core-9 binds, checker traversal wired, conversion leg live (review applied)

juanmicrosoft:
## Summary

B2 of the #762 binder rebuild (scoping doc row 2): the core-9 family binds structurally, and the F-2 conversion leg activates as the amendment staged.

- **Eight real binders**: `Some`/`Ok`/`Err` (composed `Option<T>`/`Result<T,E>` type strings, unknowable parameters as explicit placeholders per D3), `ExpressionCall` (target + arguments — the deep-payoff pin shows a division nested in an argument is analyzable again), anonymous objects, records (bound *through* the broken-`Accept` `FieldAssignmentNode` by property access — the #762 item-8 hazard handled without visitor dispatch), `with` (target's type), `throw` (`NEVER`). `SelfRefNode` stays explicitly incomplete with F-1 dormancy registered in-code — the "core row fully live" exit criterion carries that stated qualification.
- **Instrument, both legs**: in-repo ratchet **90 → 79**; the **conversion leg's first measurement — ~11.8% incomplete on converted C# (1,450/12,307 over 305 files) vs ~1.8% in-repo** — quantifies the 6× C#-shape concentration in unbound families, which is the empirical justification for the family ordering. In-process (~6s), no caching machinery needed; skips where submodules are absent.
- **Checker-delta disclosure** (suite mechanism): exactly one test changed — the fallback-opacity pin's probe swapped to Tier-B `AddressOfNode` because its old probe (`Some`) is now bound. That failure occurring at all is the mechanism working.
- Freeze amendment updated: conversion-leg activation + first numbers on the record; regen writers race-proofed.

## Verification

Nine new family tests (bound-node type, composed type strings, children retained, no `Calor0259`; SelfRef pins its dormancy reason). Full suite **8,308/0**.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,6 +33,12 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # #762 B2 (review Major 2): the F-2 conversion-leg ratchet
+          # (BinderIncompleteRatchetTests.ConversionLeg_*) skips without the corpus
+          # submodules — without this, leg 2's equality ratchet is enforced NOWHERE
+          # in CI and a converter/binder regression moving its counts fails no pipeline.
+          submodules: recursive
 
       - name: Setup .NET
         uses: actions/setup-dotnet@v4
--- a/bench/phase0-agent-native/binder-incomplete-baseline.json
+++ b/bench/phase0-agent-native/binder-incomplete-baseline.json
@@ -1,7 +1,13 @@
 {
-  "IncompleteCount": 90,
+  "IncompleteCount": 79,
   "ParsedFiles": 443,
   "ParseFailures": 9,
-  "ExpressionsBound": 4286,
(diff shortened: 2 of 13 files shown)


juanmicrosoft:
Review round applied (verdict 80%). The CRITICAL was the find of the B-series so far: **the headline claim was false end-to-end** — checkers' closed traversal switches skipped the new nodes' children, so a nested division produced no finding. Fixed at the root of the class: `BoundChildren.Of()` as the shared default arm of all four traversal switches (family PRs extend one place), with an end-to-end pin producing the real Calor0920 (stash-verified to discriminate). Leg 2 is now CI-enforced (submodules on the test job — the review proved the ratchet ran nowhere). The 59 split per-mode: **all 59 are converter output-parse failures** (#903, three clusters), each mode ratcheted separately. Ordering honesty on the record: the breakdown validates the family *set*; the conversion-leg mass concentrates in B6 (lambda = 46%), and B3's precedence is checker payoff. Minors applied (spans on BoundNamedValue, shared scope constant, dict-comment warning, friendly baseline path). Suite 8,310/0.

Issue #786, opened 2026-07-24
P1: Rebuild bug-pattern analyzers on typed CFG/SSA semantics


juanmicrosoft:
## Summary

Replace heuristic division, overflow, out-of-bounds, off-by-one, and null-dereference checks with typed CFG/SSA reasoning and adversarial runtime-oracle tests.

## Evidence and impact

- Division analysis treats an initial nonzero value as permanent and misses later assignment to zero.
- Else branches do not always receive the negated path condition.
- Off-by-one detection reports “array access” for any loop-variable reference.
- Index analysis lacks a real bound array-access node, string-matches call names, checks mainly negative values, and treats partial checks as full bounds protection.
- Overflow shortcuts miss obvious target-type overflow and mishandle unsigned/MIN÷-1 cases.
- Null analysis recognizes `.unwrap` by target-name suffix rather than actual option/null semantics and does not invalidate checked state after assignment.
- Tests include vacuous patterns such as `hasDiagnostic || true` and `Assert.NotNull`.

## Required implementation

1. Consume typed bound nodes and corrected CFG/SSA state.
2. Track strong updates/reassignments and path-condition negation.
3. Represent array/index/length/null/option operations explicitly.
4. Prove off-by-one only when a real bounded access is connected to the loop induction variable.
5. Model target-type arithmetic, promotions, checked context, signedness, and exceptional edge cases.
6. Remove name/substr-based safety conclusions.
7. Return explicit `unknown/incomplete` when required semantic information is absent.
8. Separate verified findings from heuristic hints in diagnostic codes/severity.

## Required regression coverage

- Positive and negative cases for every checker.
- Reassignment invalidating earlier facts.
- Equivalent safe/unsafe branch pairs.
- Real versus unrelated loop-variable references.
- Negative, upper-bound, multidimensional, slice/range, and string indexing.
- Numeric boundaries for every integral type.
- Null/option state changes, aliases, and repeated checks.
- Runtime oracle corpus that executes generated C# and compares actual failure behavior.

## Success metrics / definition of done

- All confirmed false-negative reproductions are detected.
- All confirmed fabricated-warning reproductions are silent.
- The adversarial positive/negative corpus reaches the project’s documented precision/recall thresholds, with zero known P0 false-safety cases.
- No production checker uses target-name substring matching as proof of semantics.
- No behavioral test uses `|| true`, `Count >= 0`, or existence-only assertions.
- Incomplete analysis is surfaced explicitly rather than treated as safe.

[...]

PR #904, opened 2026-08-10
feat(binder): #762 B3 — arrays/collections bind, retention complete (ratchet 79→51, conv 1450→1257)

juanmicrosoft:
## Summary

B3 of the #762 binder rebuild (scoping doc row 3): the thirteen arrays/indexes + collections classes bind structurally.

- **All AST properties retained** per the B2 review's standard — sizes, initializers, dimension lists, optional range bounds, dictionary entries (as span-carrying `BoundPair` records), collection-name metadata matching the AST shape.
- **Composed type strings**: `T[]`/`T[,]` with element-type derivation on access, `List<T>`/`Set<T>`/`Dictionary<K,V>`, INT/BOOL for counts and membership — 0.13 string types, explicit and non-null per D3.
- **`BoundChildren.Of()` extended** per the family contract (one place; all four traversal switches follow), per-node enumeration locked by unit test, and the **end-to-end pin**: a `/0` nested in a `§LIST` initializer produces the real `Calor0920` through the full pipeline.
- **Ratchet movement, both legs**: in-repo **79 → 51**; conversion leg **1,450 → 1,257** (−193 — matching the B2 review's B3 estimate of ~200 to the digit's neighborhood). Fractions now 1.2% in-repo / 9.8% converted.
- **Checker-delta disclosure**: zero pre-existing tests moved. Visibility-vs-consumption honesty in the test file itself — the index-OOB checker keys on call-target strings until #786, so this PR delivers traversal visibility for the array family, not an OOB-findings delta, and says so.

## Verification

Nine family tests (bound types, composed type strings, children retention, no `Calor0259`, `BoundChildren` coverage, e2e finding). Full suite **8,319/0**.

### DIFF
--- a/bench/phase0-agent-native/binder-incomplete-baseline.json
+++ b/bench/phase0-agent-native/binder-incomplete-baseline.json
@@ -1,12 +1,12 @@
 {
-  "IncompleteCount": 79,
+  "IncompleteCount": 51,
   "ParsedFiles": 443,
   "ParseFailures": 9,
-  "ExpressionsBound": 4299,
+  "ExpressionsBound": 4385,
   "Scope": "both F-2 legs; conversion leg skips when corpus submodules are absent",
   "Conversion": {
-    "Incomplete": 1450,
-    "ExpressionsBound": 12307,
+    "Incomplete": 1257,
+    "ExpressionsBound": 12766,
     "ConvertedAndBound": 305,
     "ConvertExceptions": 0,
     "EmptyOutput": 0,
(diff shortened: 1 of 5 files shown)


juanmicrosoft:
Review round applied (verdict 78%; every quantitative claim reproduced to the digit, including the −193 decomposition). All four retention gaps fixed — **ContainsMode carried** (the CRITICAL: three distinct operations), Id on all five creation nodes, **row structure retained** on multi-dim (flattening only inside `BoundChildren.Of`), Attributes noted; vocabulary aligned with the parser's own spellings (`HashSet<T>`, `Dictionary<K,V>` no-space, composed tuple types) so a §SET bind's variable and initializer no longer disagree; **`ContainsArrayAccess` now recognizes the structural nodes it was named for** with the stale placeholder comment corrected; jagged derivation fixed via trailing-bracket slicing; the e2e pin is span-anchored. Ratchet counts unchanged (retention adds no incompleteness). Suite 8,321/0.

PR #905, opened 2026-08-10
feat(binder): #762 B4 — string family binds, per-op typed results (ratchet 51→28, conv 1257→982)

juanmicrosoft:
## Summary

B4 of the #762 binder rebuild (scoping doc row 4): the four string-family classes bind with **per-operation derived result types** — the first family where the type strings carry real information rather than placeholders.

- `BoundStringOperation`: 11 distinct per-op result mappings (INT/BOOL/`str[]`/OBJECT/STRING) pinned by theory tests; **`ComparisonMode` retained** — the B3 `ContainsMode` lesson applied up front.
- `BoundInterpolatedString`: part order + **format specifiers and alignment clauses retained**; `BoundChildren.Of` enumerates expression holes only.
- `BoundStringBuilderOperation` / `BoundCharOperation`: per-op types using existing vocabulary (`StringBuilder` = the effect-resolution spelling; `CHAR` = AttributeHelper's canonical form) — no invented spellings this round.
- Span-anchored e2e pin: `/0` nested in a string-op argument produces the real `Calor0920` at the right line. (Honest note: the first pin attempt used an interpolation hole and discovered Calor interpolation doesn't parse lisp expressions in holes — the pin nests through `(str …)` instead.)

## Ratchet movement

In-repo **51 → 28**; conversion leg **1,257 → 982** (−275 — the B2 breakdown's B4 estimate was 274). Fractions now **0.63% in-repo / 7.5% converted**. Remaining conversion-leg mass is exactly the B5 (~74) and B6 (~881-minus-newly-exposed) families, as scheduled.

## Verification

23 family tests; zero pre-existing tests moved (checker-delta disclosure); full suite **8,344/0**.

### DIFF
--- a/bench/phase0-agent-native/binder-incomplete-baseline.json
+++ b/bench/phase0-agent-native/binder-incomplete-baseline.json
@@ -1,12 +1,12 @@
 {
-  "IncompleteCount": 51,
+  "IncompleteCount": 28,
   "ParsedFiles": 443,
   "ParseFailures": 9,
-  "ExpressionsBound": 4385,
+  "ExpressionsBound": 4415,
   "Scope": "both F-2 legs; conversion leg skips when corpus submodules are absent",
   "Conversion": {
-    "Incomplete": 1257,
-    "ExpressionsBound": 12766,
+    "Incomplete": 982,
+    "ExpressionsBound": 13179,
     "ConvertedAndBound": 305,
     "ConvertExceptions": 0,
     "EmptyOutput": 0,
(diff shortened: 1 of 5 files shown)


juanmicrosoft:
Review verdict: **MERGE-CLEAN (92%)** — the series' first, with every claim reproduced from scratch (both legs bit-exact; the −275 decomposed as −191 StringOp −84 CharOp −2 SB +2 newly-exposed NullCoalesce; every type mapping verified against the emitter; the e2e pin proven to fail on main's binder). Applied the minors in-PR (loud-fail interpolation fallback arm, RegexMatch collision note); recorded the **vocabulary-normalization decision in the scoping doc due in B5** (Major 1 — the bifurcation can't be fixed inside B4 since it matches both precedents); filed **#906** for the review's most interesting discovery: `${}`-embedded expressions are invisible to the binder AND the F-2 instrument — a denominator blind spot no ratchet counts.

PR #907, opened 2026-08-11
feat(binder): #762 B5 — semantic repairs: decimals, casts, type tests (conv 982→904)

juanmicrosoft:
## Summary

B5 of the #762 binder rebuild (scoping doc row 5) — the semantic-repair family. The three #762 evidence bullets die here:

- **Decimals keep full precision** (`BoundDecimalLiteral`): the pin uses a 28-digit value and asserts the old `(double)` downcast *would* have lost it — proving the defect existed, not merely that the fix works.
- **Casts carry their target type** (`BoundConversionExpression`, Cast + As): the old arm returned the operand, so a cast's static type was whatever the operand claimed.
- **`is`/patterns are real type tests** (`BoundTypeTest`, operand + pattern variable retained): the old arms returned **literal `true`** — a branch-folding hazard for constant-aware checkers.
- `typeof` binds; `TypeOp`'s fallback arm fails loud (enum-complete).

**The vocabulary-normalization decision (B4 review Major 1, due here) is resolved as documented deferral**: the two-layer TypeName reality is stated on `BoundExpression.TypeName` itself — canonical literal families, parser-surface composed forms, never equality-compared — with unification owned by 0.14's typed representation (roadmap §3.2), because a string-level fix now would either break the B3 parser-agreement or churn every family twice.

## Ratchet & disclosure

In-repo **unchanged at 28** — this family was bound-but-*wrong*, not incomplete; repairs don't move `Calor0259` counts, and saying so is the honest reading of the instrument. Conversion **982 → 904** (−78 = exactly the `TypeOf` clearance). Checker-delta: **one** pre-existing test moved — `DecimalLiteral_NotMisparsedAsZero`, which had pinned the downcast defect's implementation shape (the golden-pins-the-defect pattern, again).

## Verification

Eight family tests incl. the span-anchored e2e pin (`/0` inside a cast operand → real `Calor0920`). Full suite **8,352/0**.

### DIFF
--- a/bench/phase0-agent-native/binder-incomplete-baseline.json
+++ b/bench/phase0-agent-native/binder-incomplete-baseline.json
@@ -5,7 +5,7 @@
   "ExpressionsBound": 4415,
   "Scope": "both F-2 legs; conversion leg skips when corpus submodules are absent",
   "Conversion": {
-    "Incomplete": 982,
+    "Incomplete": 904,
     "ExpressionsBound": 13179,
     "ConvertedAndBound": 305,
     "ConvertExceptions": 0,
--- a/docs/plans/binder-rebuild-762-scoping.md
+++ b/docs/plans/binder-rebuild-762-scoping.md
@@ -164,9 +164,12 @@ items 1–3 (B1–B8 families), 4 (B5), 5–6 (B8), 7 (B8), 8 (B1 expression-hal
   vocabulary is bifurcated — scalars use canonical forms ("STRING") while composed types use
   surface forms ("str[]", "i32[]"), so array-element derivation yields "str" where scalar
   string expressions say "STRING". Latent (no consumer equality-compares bound TypeNames —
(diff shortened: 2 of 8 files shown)


juanmicrosoft:
Review round applied (verdict 72%). Both divisor regressions repaired with **stash-verified discriminating pins** — and the pins themselves had a near-miss worth recording: the first drafts silently tested all-findings mode (`BugPatternOptions.ReportOnlyVerified` defaults false; the CLI default is true), so the cast pin passed without its fix. Tightened to the CLI default + severity discrimination; both now fail without the fixes. Pattern variables are **declared in scope** (retained-but-unusable wasn't good enough for this node's family PR); `TYPE` spelling corrected; §5 records the deferral decision (per the reviewer's position that documented deferral legitimately discharges B4's ask); disclosure corrected to one-test-moved-plus-two-review-caught-regressions. Suite 8,355/0.

PR #908, opened 2026-08-11
#762 B6: bind the control-value family (null-coalesce, null-conditional, match, lambda, await)

juanmicrosoft:
## Summary

Sixth family PR of the #762 binder rebuild (B6 of B1–B8, per the [scoping doc](../blob/main/docs/plans/binder-rebuild-762-scoping.md)). Binds the five control-value classes — the conversion-leg payload; lambda alone accounted for 46% of the original conversion-leg incomplete mass.

## The deferred-evaluation decision

This family is the first whose children are **conditionally executed** (`??` fallback, match guards, lambda bodies). Two-part decision, pinned by tests:

- `BoundChildren.Of()` **keeps seeing everything** — value-safety traversals deliberately walk deferred subtrees, because a `/ 10 0` in a `??` fallback is a real latent bug (the question is *when* it executes, not *whether* it's wrong). E2E pin: `§R (?? x (/ 10 0))` produces Calor0920 at the right span with `ReportOnlyVerified = true`.
- New `BoundChildren.DeferredOf()` identifies the not-necessarily-executed subset for future occurrence-sensitive analyses (nothing consumes it yet; each family test pins its contents).

## Nodes

| Bound node | Shape retained | TypeName |
|---|---|---|
| `BoundNullCoalesce` | left, right | left's |
| `BoundNullConditional` | target, member name | `OBJECT` |
| `BoundMatchExpression` | scrutinee, cases (Pattern AST + Guard + Body), Id | `OBJECT` |
| `BoundLambda` | Id, params, IsAsync, IsStatic, expr/statement body | `OBJECT` |
| `BoundAwaitExpression` | awaited, ConfigureAwait | unwraps `Task<T>`→`T`, `Task`→`VOID`, else `OBJECT` |

`BoundMatchExpressionCase` is named to avoid colliding with the existing match-*statement* case class.

## Scoping

`BindLambda` opens a child scope and declares the lambda parameters — bodies now bind for real (conversion-leg ExpressionsBound 13,179 → 17,661) and parameter names do not leak outward (leak-probe test: `y` binds inside, Calor0201 outside).

## Ratchet movement

| Leg | Before | After |
|---|---|---|
| In-repo incomplete | 28 | **1** |
| Conversion incomplete | 904 | **21** |

Remaining mass is B7 (quantifiers) + Tier B.

The LSP Info-severity instrument test used `??` as its unbound fixture — re-pointed to a B7 quantifier per its own fixture-migration note.

Part of #762.

### DIFF
--- a/bench/phase0-agent-native/binder-incomplete-baseline.json
+++ b/bench/phase0-agent-native/binder-incomplete-baseline.json
@@ -1,12 +1,12 @@
 {
-  "IncompleteCount": 28,
+  "IncompleteCount": 1,
   "ParsedFiles": 443,
   "ParseFailures": 9,
-  "ExpressionsBound": 4415,
+  "ExpressionsBound": 4567,
   "Scope": "both F-2 legs; conversion leg skips when corpus submodules are absent",
   "Conversion": {
-    "Incomplete": 904,
-    "ExpressionsBound": 13179,
+    "Incomplete": 21,
+    "ExpressionsBound": 17661,
     "ConvertedAndBound": 305,
     "ConvertExceptions": 0,
     "EmptyOutput": 0,
--- a/src/Calor.Compiler/Analysis/Dataflow/Analyses/UninitializedVariables.cs
+++ b/src/Calor.Compiler/Analysis/Dataflow/Analyses/UninitializedVariables.cs
@@ -94,6 +94,12 @@ public sealed class UninitializedVariablesAnalysis
             {
                 foreach (var v in BoundNodeHelpers.GetUsedVariables(block.BranchCondition))
                 {
+                    // Parameters are initialized by their caller: function parameters
+                    // are seeded Initialized at entry, and a LAMBDA parameter (a scope
+                    // this name-keyed analysis cannot see) is always initialized when
+                    // its body runs — flagging it is wrong on any execution (#908 F2).
+                    if (v.IsParameter)
+                        continue;
                     var state = currentFacts.GetState(v.Name);
                     if (state != InitializationState.Initialized)
                     {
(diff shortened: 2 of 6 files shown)


juanmicrosoft:
**Adversarial review complete** (lens rotated per policy: deferred-evaluation semantic fidelity + scope lifecycle — B2 used e2e reachability, B5 used checker-verdict/pin-discrimination). Verdict was FINDINGS-TO-APPLY; applied in a7d17af8:

- **F1 (blocker, confirmed)**: expression-match arms shared the enclosing scope — false `Calor0201` across arms, arm locals leaking outward, variable-pattern captures erroring. Fixed with per-arm child scopes mirroring `BindMatchStatement`, plus capture declaration.
- **F2 (blocker, confirmed)**: every expression-body lambda produced an Error-severity `Calor0900` on its own parameter (name-keyed analysis defaults unknown names to Uninitialized). Fixed by skipping `IsParameter` symbols in use detection.
- **F3**: match-arm bodies are in neither `Of()` nor `DeferredOf()` — docstrings corrected and the blindness pinned as current-state (gap owned by #786).
- **F4**: `BoundLambda` now retains `Effects` (AST-retained until 0.15 effect-rows).
- **F5**: duplicate lambda parameter names now report `Calor0201`.

**Pin discrimination verified**: armed the five pre-fix behaviors and watched exactly the five new pins fail (8 originals green), then restored and got 13/13.

Notes not applied here: F6 (wrapper-nested division double-report + over-strong `Calor0926` suggestion for deferred divisions — pre-existing, filed as a follow-up issue), F7 (await unwrap degrade cases confirmed harmless: no consumer equality-compares TypeName).

Review also confirmed the LSP fixture re-point discriminates: the `forall` fixture emits exactly one Info `Calor0259` and fails the moment B7 binds quantifiers.

Issue #909, opened 2026-08-11
DivisionByZeroChecker double-reports divisions nested under bound wrapper nodes; Calor0926 suggests over-strong contracts for deferred divisions


juanmicrosoft:
Found by the #908 (B6) adversarial review — F6, pre-existing pattern that B6 widens.

**Double-reporting**: `DivisionByZeroChecker.CheckExpression` runs `ContainsDivision` at every recursion level, so a division nested under any traversed wrapper is checked at the wrapper AND at the division node itself. Probe: `(?? x (/ 10 y))` produces 2× Calor0920 vs 1× for plain `(/ 10 y)`. Any B2–B5 wrapper reproduces this; B6's `??`/match-guard wrappers widen it.

**Over-strong suggestion**: the Calor0926 "add §Q (!= y 0)" suggester now fires for divisions that only execute conditionally (e.g. the `??` fallback) — the suggested precondition is stronger than the code requires. A deferred-aware suggester should consult `BoundChildren.DeferredOf()` (introduced in #908) before proposing function-level contracts for conditionally-executed divisions.

Related: the #786 checker re-platform is the natural home for both (single-visit traversal + deferral awareness).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

juanmicrosoft:
The #910 (B7) review found the same mechanism produces a **3×** Calor0920 for a division inside a quantifier body (baseline non-quantifier /0 already reports 2×) — each Of()-traversed wrapper level re-runs ContainsDivision, so every new bound family adds a copy. Scale grows with B-series depth; still cosmetic (pins use Assert.Contains), still owned by the #786 single-visit re-platform.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

juanmicrosoft:
Probed on main at cc62c4ac (#900). Result: **the double-reporting half is fixed; the over-strong Calor0926 half is NOT** — leaving open with narrowed scope.

**Double-reporting — FIXED.** `DivisionByZeroChecker.CheckExpression` was rewritten in cc62c4ac to check only when the current node *is* a division (`BoundBinaryExpression` Divide/Modulo) and then recurse via `BoundNodeHelpers.GetChildExpressions` — the per-level `ContainsDivision` re-scan that caused wrapper-level duplicate reports is gone. Probe (`EnableVerificationAnalyses`, `ReportOnlyVerified=true`):

| Case | Calor0920 count |
|---|---|
[...]

PR #910, opened 2026-08-11
#762 B7: bind the quantifier family (forall, exists, implication)

juanmicrosoft:
## Summary

Seventh family PR of the #762 binder rebuild (B7 of B1–B8, per the [scoping doc](../blob/main/docs/plans/binder-rebuild-762-scoping.md)). Binds the three quantifier classes: `ForallExpressionNode`, `ExistsExpressionNode`, `ImplicationExpressionNode`.

## The family's checker contract, scoped precisely (review round applied)

Quantifiers are SPEC expressions. **Contract verification is untouched**: contracts (§Q/§S/§IV) are never bound at all — `BindFunction` binds parameters and body statements only — and the contract Z3 pipeline consumes quantifier **AST** (`ExpressionSimplifier`). A contract quantifier cannot produce different diagnostics post-B7 (review-confirmed by experiment, including a §Q with duplicate bound vars + undefined ref + /0: zero binder diagnostics on both analyze and LSP paths). The untouched 375-test `Calor.Verification.Tests` suite is that contract's baseline.

Two deliberate interactions, disclosed: the **bug-pattern Z3 path** now consumes bound quantifier bodies — that's the point of binding them, and the division pin proves it — and **§PROOF** (a bound spec position) now surfaces real scope diagnostics (e.g. Calor0201 for duplicate bound vars) on the LSP live bag instead of Info Calor0259.

## Nodes

| Bound node | Shape | TypeName |
|---|---|---|
| `BoundForallExpression` | bound variables (as symbols), body | `BOOL` |
| `BoundExistsExpression` | same | `BOOL` |
| `BoundImplicationExpression` | antecedent, consequent | `BOOL` |

## Scoping (applies the B6 review lessons up front)

- Quantifier variables are declared in a **child scope**, marked `isParameter` — bound *by* the quantifier, so the name-keyed uninitialized-variables analysis can never flag them (the B6 F2 rule covers them with zero new analysis code). Review confirmed all three repo-wide `IsParameter` consumers are safe for these symbols.
- No leak outward (leak-probe pin); duplicate bound-variable names report Calor0201 (function-parameter parity).
- `Of()`: bodies visible — a `/ 10 0` inside a forall body is a bug in the spec; e2e pin proves Calor0920 fires at the right span. `DeferredOf()`: quantifier bodies (empty domain → zero evaluations) and the implication consequent (`!a || b` short-circuits).
- Parsed-source e2e pins for **all three** spellings: `(forall …)`, `(exists …)`, `(-> …)` — review confirmed `(-> a b)` is the only implication spelling and cannot collide with the return arrow.

## Ratchet: zero movement — because contracts are never bound

Corrected per review: quantifiers **do** occur in the in-repo corpus (`07_quantifiers.calr` under two roots) — but exclusively in §Q positions, and contracts are never bound. So B7's binding is exercised by expression-position and §PROOF-position quantifiers only, which outside this PR's tests occur nowhere in the repo today. The "family binds" exit criterion is met; the practical payload arrives when quantifiers appear in executable positions (and when B8's routing/closure work broadens what analyze surfaces).

Remainder-by-class (review-reproduced byte-for-byte):

| Leg | Remaining incomplete | Classes |
|---|---|---|
| In-repo | 1 | `SizeOfNode` (Tier B, unsafe family) |
| Conversion | 21 | 17× `StackAllocNode` (Tier B, unsafe), 4× `RawCSharpExpressionNode` (B8 interop structural binding) |

**Post-B7, every remaining incomplete on both legs is Tier B or B8-owned** — the B8 "zero Tier A incomplete" gate-1 criterion is now within B8's own scope.

The LSP Info-instrument fixture moves to its third construct (`??` → `forall` → `§SIZEOF{i32}`), which holds until B8's closure decision owns re-pointing or retiring it.

Part of #762.

### DIFF
--- a/src/Calor.Compiler/Binding/Binder.cs
+++ b/src/Calor.Compiler/Binding/Binder.cs
@@ -515,9 +515,30 @@ public sealed class Binder
             [typeof(NullConditionalNode)] = (b, e) =>
                 { var n = (NullConditionalNode)e; return new BoundNullConditional(n.Span, b.BindExpression(n.Target), n.MemberName); },
             [typeof(MatchExpressionNode)] = (b, e) => b.BindMatchExpression((MatchExpressionNode)e),
-            [typeof(LambdaExpressionNode)] = (b, e) => b.BindLambda((LambdaExpressionNode)e),
+            [typeof(LambdaExpressionNode)] = (b, e) =>
+                b.BindLambda((LambdaExpressionNode)e),
             [typeof(AwaitExpressionNode)] = (b, e) =>
                 { var n = (AwaitExpressionNode)e; return new BoundAwaitExpression(n.Span, b.BindExpression(n.Awaited), n.ConfigureAwait); },
+            // #762 B7 — quantifier family (spec expressions; Z3 consumes their AST,
+            // not these bound nodes — binding is for value-safety visibility only).
+            [typeof(ForallExpressionNode)] = (b, e) =>
+                {
+                    var n = (ForallExpressionNode)e;
+                    var (vars, body) = b.BindQuantifier(n.BoundVariables, n.Body);
+                    return new BoundForallExpression(n.Span, vars, body);
+                },
+            [typeof(ExistsExpressionNode)] = (b, e) =>
+                {
+                    var n = (ExistsExpressionNode)e;
+                    var (vars, body) = b.BindQuantifier(n.BoundVariables, n.Body);
+                    return new BoundExistsExpression(n.Span, vars, body);
+                },
+            [typeof(ImplicationExpressionNode)] = (b, e) =>
+                {
+                    var n = (ImplicationExpressionNode)e;
+                    return new BoundImplicationExpression(n.Span,
+                        b.BindExpression(n.Antecedent), b.BindExpression(n.Consequent));
+                },
         };
 
         // Every remaining concrete ExpressionNode subclass dispatches to BindIncomplete.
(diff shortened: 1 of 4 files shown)


juanmicrosoft:
**Adversarial review complete** (lens rotated: adversarial claim-verification — B2 e2e reachability, B5 checker verdicts/pins, B6 deferral fidelity/scope lifecycle). Verdict: FINDINGS-TO-APPLY, no code defect. Applied in dca51b48:

- **Ratchet rationale corrected** (the review's top finding): zero movement is NOT because quantifiers are absent from the corpus — `07_quantifiers.calr` has all three forms — but because they live exclusively in §Q and **contracts are never bound**. PR body rewritten to say so, including the honest consequence: the family's binding is exercised nowhere outside its own tests today.
- **e2e pins added for `exists` and `(-> …)`** — previously only forall reached the binder from parsed source in the suite.
- **Z3 claim scoped**: contract verification untouched (review-confirmed by experiment); the bug-pattern Z3 path deliberately consumes bound quantifier bodies; §PROOF quantifiers now surface real scope diagnostics on the LSP live bag.
- Routing-vacuity comment added (analyze routes only Calor0259 from the binder bag — the real scope guards are the live-bag unit tests); line-comment nit fixed.
- Confirmed clean by the review: isParameter leakage (all 3 repo-wide consumers safe), §SIZEOF fixture parses outside unsafe and discriminates both ways, remainder-by-class table reproduced byte-for-byte, 3 of 4 runnable pins fail against the pre-B7 binder (the leak probe is a guard pin, disclosed).
- Follow-up noted on #909: Calor0920 now triples inside quantifier bodies (same #786-owned dedup root cause).

PR #911, opened 2026-08-11
#762 B8: closure — interop, item-8, overload sets, expression-start unification, seed suite, gate 1

juanmicrosoft:
## Summary

Final PR of the #762 binder rebuild (B8 of B1–B8, per the [scoping doc](../blob/main/docs/plans/binder-rebuild-762-scoping.md)). Four commits, one per workstream cluster. Several designs are deliberately ported from PR #900 (the parallel full implementation) with attribution — the ExpressionParsers table, the `ParseAttributes(maxGroups)` companion fix, the item-8 visitor implementations, the DispatchProxy architecture test, and the token-context matrix test.

## What closes here

**Interop (Tier A closure).** `RawCSharpExpressionNode` / `FallbackExpressionNode` bind with verbatim content + stable `OBJECT` type + `IsInterop` marker (F-1's interop clause). Conversion-leg incomplete 21 → 17; **Tier A incomplete is now ZERO on both F-2 legs** — §2.5 **gate 1 measured green** (annotated in the roadmap; residuals published: 1 `SizeOfNode` in-repo, 17 `StackAllocNode` conversion, all Tier B).

**Item 8, full scope.** The seven non-expression no-op-Accept classes get real Visit dispatch in both interfaces and all five implementers; `GenericTypeNode` promotes to Tier A (binds structurally); `KeywordArgNode` is **reclassified out of the AST** (parser-internal value object — the no-op-Accept null-injection hazard is now unrepresentable, not merely guarded). The AstNode-wide architecture test (DispatchProxy + `GetUninitializedObject`) makes any future no-op Accept a CI failure. F-1 amendment recorded: denominator 61 → 60 (56 Tier A / 4 Tier B), subtractive-with-supersession argument in the freeze doc.

**D5 overload sets (items 5–6).** Top-level functions register in overload sets (previously a same-name second declaration *silently vanished* and every call resolved to the first). New: Calor0206 duplicate-signature Error at the declaration; Calor0207 arity-tie / Calor0208 no-arity-match Warnings at the call — resolution proceeds with the first declaration for bound-tree continuity, flagged, never silent. Order-independence pinned.

**Item 7 expression-start unification.** `ParseExpressionCore`'s 47-arm switch and `IsExpressionStart`'s hand-maintained list are now two projections of ONE static table. The measured pre-B8 disagreement: exactly `OpenBrace` (dispatchable, not start-classified) — brace collection initializers silently failed to parse in start-gated contexts; fixing it required the `ParseAttributes(maxGroups: 1)` companion so `§B{x} {1, 2}` doesn't eat the initializer as a header group. Token-context matrix: every registered token × return/bind/argument/nested (188 cases) + a completeness fact tying the case table to the registry.

**Diagnostic-seed suite (#762 required coverage).** A `/ 10 0` seed inside 13 wrappers proven to reach the division checker — including the Tier-B retained-children rows, which caught a real bug in this PR's own first draft: `BoundIncompleteExpression` matched checkers' `BoundCallExpression` arms before any `BoundChildren` default, so retained children were invisible. Four traversals now walk `RetainedChildren` explicitly. Honest exclusions documented (match-arm bodies / lambda statement bodies = #786 gap).

**Tier-B child extractors (D2).** The four unsafe classes stay incomplete (Calor0259 unchanged — the gate bar does not move) but their subtrees bind and ride on `BoundIncompleteExpression.RetainedChildren`, deferred-marked.

**Severity promotion (scoping doc §5's B8 decision).** Calor0259: Info → **Warning**. Tier A incomplete is zero, so a warning now genuinely means "analysis does not cover this construct" — only Tier B unsafe residuals emit it.

## #762 DoD checklist

| DoD bullet | Status |
|---|---|
| 100% of accepted expression kinds structurally bind | 56/56 Tier A bind; 4 Tier B residuals retain children + diagnostic (item-3 end state) |
| No silent zero-child `<unsupported:...>` | Calor0259 Warning + RetainedChildren; ratchet enforces |
| Stable type + retained evaluated children | Per-family bound nodes B2–B8 |
| Every concrete visitable AstNode dispatches exactly once | ArchitectureTests (AstNode-wide, this PR) |
| Overload resolution never silently falls back | Calor0206/0207/0208 (this PR) |
| Binder completeness measured in CI, cannot regress | Two-leg ratchet + dispatch completeness + matrix tests |

SymbolId population (item 9) was plumbed in B1; population-at-scale and consumers remain the roadmap's gates 3/4 work, not B8's.

Full suite: 7,607 tests green.

Part of #762 — closes the B1–B8 execution.

### DIFF
--- a/bench/phase0-agent-native/binder-incomplete-baseline.json
+++ b/bench/phase0-agent-native/binder-incomplete-baseline.json
@@ -5,8 +5,8 @@
   "ExpressionsBound": 4567,
   "Scope": "both F-2 legs; conversion leg skips when corpus submodules are absent",
   "Conversion": {
-    "Incomplete": 21,
-    "ExpressionsBound": 17661,
+    "Incomplete": 17,
+    "ExpressionsBound": 17678,
     "ConvertedAndBound": 305,
     "ConvertExceptions": 0,
     "EmptyOutput": 0,
--- a/docs/plans/roadmap-v0.13-v0.15.md
+++ b/docs/plans/roadmap-v0.13-v0.15.md
@@ -167,7 +167,9 @@ gates 1, 4, 5, 6, 7, and gate 2's diagnostics leg.
    strictly better instrument); the residual incomplete-fraction outside those classes is
    **published per release, reported-not-adjudicated** (the M-S2/PP-L4 route). The escape
    category is thereby bounded: the gate cannot be passed by marking registered constructs
-   "incomplete".
+   "incomplete". **Status: measured green by the B8 closure PR (2026-08-10)** — zero Tier A
+   `Calor0259` on both F-2 legs; residuals published in the F-1 B8 amendment (in-repo 1
+   `SizeOfNode`, conversion 17 `StackAllocNode`, all Tier B).
 2. **Full-vs-incremental identity**: byte-identical diagnostics, index contents, and review packets
    (after canonical ordering) across a **registered edit-script corpus** that includes the #883
    reproduction — plus an **incrementality witness** (the incremental path demonstrably reuses the
--- a/docs/plans/v0.13-freeze-registrations.md
+++ b/docs/plans/v0.13-freeze-registrations.md
@@ -76,9 +76,13 @@ amendment rules).
 
 ### F-1 amendment — B8 item-8 dispositions, interop closure, and gate-1 evaluation (2026-08-10)
 
-- **GenericTypeNode → Tier A (additive promotion).** It is a real expression-position type
-  reference with real Accept dispatch; it now binds structurally
-  (`BoundGenericTypeExpression`, composed parser-surface TypeName).
+- **GenericTypeNode → Tier A (additive promotion), DORMANT** (disclosure per the
+  SelfRefNode precedent, added by the #911 review): the binder is real
+  (`BoundGenericTypeExpression`, composed parser-surface TypeName) but the class is
+  currently **unreachable from source text** — the legacy `§G` lexer keyword was removed,
+  so the only producer (`Parser.ParseGenericType`) has no live token route, and the
+  matrix test's Generic row runs on hand-built tokens (labeled "synthetic"). If a
+  construction path returns, the promotion carries F-1's dormant-rule corpus obligation.
 - **KeywordArgNode → out of the denominator (subtractive, with supersession).** #762 item 8's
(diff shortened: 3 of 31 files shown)


juanmicrosoft:
**Adversarial review complete** (lens rotated: behavioral blast radius of the parser/policy changes — B2 e2e reachability, B5 verdicts/pins, B6 deferral/scope, B7 claim verification). Verdict FINDINGS-TO-APPLY, no blocker; all applied in a9112e78:

- **F1 (confirmed)**: the Calor0259 Info→Warning promotion was silently inert on the CLI/analyze/MCP surface — `VerificationAnalysisPass` re-reported with hardcoded `ReportInfo`. Now propagates the binder's original severity, with a deliberate companion decision: `AnalyzeTool` excludes 0259 from `filesWithIssues` **by code** (it displays as a Warning but remains an instrument at the aggregate level). Pinned on the analyze path.
- **F2 (confirmed)**: two of the four RetainedChildren traversal arms shipped unpinned. Added an armed-revert-verified pin for `GetUsedVariables` (uninitialized read inside `§ADDR` → Calor0900); disclosed honestly that `ContainsArrayAccess`'s arm has **no reachable Tier-B consumer** until the unpatched checkers land (#912) rather than counting it proven.
- **F3**: partial-visibility disclosure — #912 filed enumerating every checker walk that still sees Tier-B subtrees as opaque; the 0259 message now names the actual boundary.
- **F4**: GenericTypeNode promotion recorded as **DORMANT** in the F-1 amendment (no lexer producer since `§G` was removed; SelfRefNode precedent; corpus obligation on revival).
- **F5**: item-8 emitter bodies labeled reference projections, and the one real divergence fixed — `Visit(EffectsNode)` emitted `§E{}` (which *means pure*) for an empty map the live path omits.
- **F6 (confirmed silent-corruption case)**: `maxGroups: 1` extended to `§C`/`§W`/`§LAM` headers — pre-fix, `§C{Sink} {1, 2} §/C` silently merged the braces into the call target with zero diagnostics. Pinned, armed-revert verified.
- Also corrected from the review's clean checks: overload policy has **zero** blast radius on the in-repo corpus (0×0206/0207/0208, byte-identical scans); the corpus-wide pre/post parse comparison is byte-identical; gate-1 reproduced independently; the KeywordArgNode supersession argument confirmed sound. The PR is 6 commits (the body's "four" predates the #897 GC fixture and this commit).

PR #914, opened 2026-08-11
#778: verification cache keys — exhaustive, semantics-versioned, solver-config-aware

### LINKED ISSUE #778: P0: Make verification cache keys exhaustive and semantics-versioned
## Summary

Make verification-cache keys exhaustive, collision-resistant, and versioned by every semantic input that can change a proof.

## Evidence and impact

The translator supports expression kinds that `ContractHasher.AppendExpression` does not serialize. Distinct string operations, field accesses, and self references can therefore collapse to the same `UNSUPPORTED:<type>` hash. Cache validity checks format and Z3 version but not compiler, binder, translator, runtime-semantics, or policy versions.

A cached `Proven` result can consequently be reused for a different contract and cause its runtime guard to be removed.

## Required implementation

1. Define an exhaustive canonical serialization for every proof-relevant AST/bound node.
2. Include node kind, type, signedness/width, operator, member identity, constants, flags, effects, preconditions, path facts, and semantic-version inputs.
3. Make unsupported node serialization include its full conservative structure; preferably do not cache unsupported outcomes.
4. Namespace cache keys by compiler version, language semantics version, binder schema, translator schema, runtime model version, solver version/configuration, and relevant verification options.
5. Cache only explicitly safe terminal outcomes.
6. Store sufficient metadata to audit why an entry matched.
7. Fail closed on malformed, partial, or unknown-version entries.
8. Add a one-command cache purge/migration strategy.

- Pairwise-distinct hashes for all concrete expression nodes and semantic flags.
- Known collision pairs for string operations, fields, and self references.
- Version/option flips that force misses.
- Corrupted and partially written cache files.
- Concurrent readers/writers and atomic save behavior.
- A proven contract followed by a distinct refuted contract with otherwise identical signatures.

- Architecture tests prove every proof-relevant node has an explicit serializer case.
- The collision corpus contains zero collisions.
- Any semantic/version/option change produces a cache miss.
- Corrupt/unknown entries can never produce `Proven`.
- The proven-then-refuted reproduction never reuses the first result.
- Cache hit/miss reasons are observable in verbose diagnostics.

## Dependencies

Depends on the typed semantic model and version definitions. Proof-based elision must remain disabled until complete.

juanmicrosoft:
## Summary

Closes the remaining #778 scope. Prior work had already fixed the three named collision classes (#824 review: string ops, field accesses, array lengths) and established the never-cache-Unsupported rule; this PR makes the remaining guarantees structural instead of incidental.

## Key exhaustiveness

The cacheable key surface is exactly `ModeledForms.ExpressionKinds` — Unsupported verdicts are never cached, so only whitelisted contracts produce cache entries. New completeness pin (`EveryModeledExpressionKind_HashesWithContent`): every whitelisted kind serializes with **content** (two distinct instances → distinct keys), with a **bidirectional** tie to the whitelist — adding a kind to `ModeledForms` without a hasher case fails the test.

Found and fixed: `SelfRefNode` (whitelisted, refinement `#`) fell to the hasher's default `UNSUPPORTED:` arm. It's contentless, so no live collision existed — it now hashes as a fixed `SELF` token.

**Defense in depth** (the issue's "preferably do not cache unsupported outcomes", made structural): the default arm now sets `SawUnhashedKind`, and `VerificationCache` refuses **both store and lookup** under a collision-unsafe key. If a future whitelist change ever outruns the hasher, the worst case is a cold re-verify, never a served collision.

## Version namespacing (cache format 1.13 → 1.14, ledger entry added)

- **`SemanticsVersion`**: entries carry `CurrentCompilerSemanticsVersion`; validity rejects mismatches. Semantics changes no longer depend solely on someone remembering the manual format-bump rule.
- **`SolverTimeoutMs`**: an `Unproven` verdict means "could not prove within the budget" — it is now reusable only when the cached budget ≥ the current one (a 5s Unproven is not served to a 30s run that might prove). Unknown budgets are conservatively non-reusable; definitive verdicts (Proven/Disproven/Assumed) remain budget-independent, which is sound. `ContractVerificationPass` passes the live timeout.

## Pin discrimination

The four new tests fail structurally without the changes: the refusal test hits the cache and finds a stored entry without the flag; the SELF pin reads `UNSUPPORTED:SelfRefNode`; the semantics/timeout tests exercise overloads that did not exist. The completeness pin's bidirectional list-equality catches drift in either direction.

Full suite: 7,614 tests green (includes all 375 verification tests).

Closes #778.

### DIFF
--- a/src/Calor.Compiler/Verification/ContractVerificationPass.cs
+++ b/src/Calor.Compiler/Verification/ContractVerificationPass.cs
@@ -53,7 +53,9 @@ public sealed class ContractVerificationPass
 
         using var ctx = Z3ContextFactory.Create();
         using var verifier = new Z3Verifier(ctx, _options.TimeoutMs);
-        using var cache = new VerificationCache(_options.CacheOptions);
+        // #778: the cache key/validity is solver-config-aware — pass the live timeout
+        // so a cached Unproven from a smaller budget is not served to a larger one.
+        using var cache = new VerificationCache(_options.CacheOptions, _options.TimeoutMs);
 
         foreach (var function in EnumerateContractBearers(module))
         {
--- a/src/Calor.Compiler/Verification/Z3/Cache/ContractHasher.cs
+++ b/src/Calor.Compiler/Verification/Z3/Cache/ContractHasher.cs
@@ -17,6 +17,7 @@ public sealed class ContractHasher
         IReadOnlyList<(string Name, string TypeName)> parameters,
         RequiresNode precondition)
     {
+        ResetUnhashedKindFlag();
         var sb = new StringBuilder();
         sb.Append("PRE:");
         AppendParameters(sb, parameters);
@@ -40,6 +41,7 @@ public sealed class ContractHasher
         EnsuresNode postcondition,
         IReadOnlyList<StatementNode>? body = null)
     {
+        ResetUnhashedKindFlag();
         var sb = new StringBuilder();
         sb.Append("POST:");
         AppendParameters(sb, parameters);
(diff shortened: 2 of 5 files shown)


juanmicrosoft:
**Adversarial review complete** (lens rotated: cache-contract soundness under adversarial inputs and concurrency). Verdict FINDINGS-TO-APPLY, no blocker; all applied in the follow-up commit:

- **F1 (confirmed, the review's best find)**: Z3 treats `timeout=0` as *infinite* — proved empirically with a probe — but the budget rule ranked it smallest, so an infinite-budget run accepted any finite-budget Unproven. Normalized (0 and `uint.MaxValue` → largest) and pinned both directions.
- **F2 (confirmed)**: the "Assumed is budget-independent" comment was wrong — Assumed stores as `Status=Unproven` (frozen D-G2.2) and is budget-gated, which the review argued is actually *correct* (Assumed can be a budget-limited downgrade of Proven via the divisor-entailment UNKNOWN path). Comment now tells the truth; refinement via stored `ProofStatus` left as a possible later optimization.
- **F3 (arm-revert confirmed)**: the lookup-side refusal guard had no discriminating pin — added one that plants a valid entry at the unhashable key's disk path and asserts refusal.
- **F4 (confirmed SDK-level collisions)**: `[("a","T"),("b","U")]` vs `[("a","T,b:U")]` hashed identically — collision resistance held only by lexer accident. All raw caller text is now length-prefixed; collision pinned; canonical-text pins updated (rides the 1.14 bump).
- **F5**: `GetCanonicalExpression` flag reset.
- Review confirmed clean: the timeout rule's direction (Unproven@5000→1000ms reuse is sound), no unguarded path to `TryGetCachedResult`/`CacheResult`, both eviction layers for pre-1.14 entries, no parallel #778 gap in KInduction/other verification components, and the completeness pin's set-equality is live. F7's point stands and is accepted: the semantics-version field is an *additional* coupling — the manual format-bump ledger discipline remains the operative protection.

PR #916, opened 2026-08-11
Z3 chain closure: drop osx-x64, upstream osx-arm64, arch-vs-RID assertion

juanmicrosoft:
## Summary

The final 0.13 MUST (roadmap §2.4 "Z3 asset chain finished"), executing the maintainer decision (2026-08-11): **drop the osx-x64 RID and document**, rather than shipping upstream's mislabeled archive (an arm64 dylib under the x64 label — Intel Macs installed successfully and *silently lost verification*) or paying for a recurring Intel source build.

## Changes

1. **osx-x64 dropped everywhere**: build-z3.yml prebuilt matrix, publish-nuget staging/packing, `Calor.Compiler`/`Calor.Tasks`/`Calor.Sdk` csproj native lists and existence gates, `inspect-sdk-nupkg.sh`, `download-z3.sh`/`.ps1`, both checksum manifests. CHANGELOG documents Intel macOS as unsupported for verification (compilation unaffected; Z3 features report "Z3 unavailable").
2. **The ARM64-macOS source-build job is deleted** — its own comment block admitted the justification had lapsed; it was the only producer of a non-reproducible asset and made the shipped osx-arm64 differ from what every ARM64-macOS developer builds against. osx-arm64 now ships the same checksum-verified upstream archive as every other RID.
3. **The held-back native arch-vs-RID assertion lands**: each prebuilt matrix leg asserts `file` output on the extracted native contains the RID's expected architecture marker (`x86-64`/`aarch64`/`arm64`/`Intel 80386`/`Aarch64`), failing the packaging job if upstream ever mislabels again — the exact check the osx-x64 defect evaded.

## Republish procedure (post-merge, per the manifest's documented rule)

Dispatch build-z3.yml once (the "one repin/republish" the roadmap names); the run prints the regenerated `.github/z3-binaries-4.15.7.sha256` block in its step summary; commit it in the same change-set. The osx-arm64 asset hash will change (source-built → upstream bytes, predictable from the pinned upstream archive); every manifest entry is then derivable from a pinned input for the first time.

Verification.Tests 375/375 and Tasks.Tests 88/88 green; full build clean; actionlint clean (no new findings).

### DIFF
--- a/.github/scripts/inspect-sdk-nupkg.sh
+++ b/.github/scripts/inspect-sdk-nupkg.sh
@@ -50,7 +50,7 @@ fi
 if [ "$ALL_RIDS" = "--all-rids" ]; then
   require "tasks/net10.0/runtimes/linux-x64/native/libz3.so"
   require "tasks/net10.0/runtimes/linux-arm64/native/libz3.so"
-  require "tasks/net10.0/runtimes/osx-x64/native/libz3.dylib"
+  # osx-x64 dropped (Z3 chain closure): Intel macOS unsupported for verification.
   require "tasks/net10.0/runtimes/osx-arm64/native/libz3.dylib"
   require "tasks/net10.0/runtimes/win-x64/native/libz3.dll"
   require "tasks/net10.0/runtimes/win-arm64/native/libz3.dll"
--- a/.github/workflows/build-z3.yml
+++ b/.github/workflows/build-z3.yml
@@ -36,115 +36,21 @@ env:
   MANAGED_DLL_ARCHIVE: x64-glibc-2.39
 
 jobs:
-  # Build Z3 from source on ARM64 macOS ONLY.
-  #
-  # ⚠ THE STATED JUSTIFICATION FOR THIS JOB NO LONGER HOLDS. It used to read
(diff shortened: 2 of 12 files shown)


juanmicrosoft:
**Adversarial review complete** (lens: release-chain integrity across the transition — 7th distinct lens in the rotation). Verdict FINDINGS-TO-APPLY, no blocker; applied in 3107d684: F1 stale-asset deletion added to the documented republish procedure (the release action never deletes unlisted assets), F2 unconditional dev-checkout purge of the dropped RID in both download scripts (provenance stamp doesn't invalidate on RID-set changes), F3 roadmap gate 7 re-oracled — the osx-x64 leg now asserts the documented loud degradation instead of a solver verdict, making the drop a tested claim.

Confirmed by the review's experiments: both transition windows fail closed (window-A pack matches the untouched osx-arm64 pin byte-for-byte; window-B is rejected by sha256 --strict with exactly one predicted delta); all six arch markers verified against the real upstream binaries on ubuntu's file(1); the upstream x64-osx dylib is byte-identical to the arm64 one — independent confirmation of the mislabeling; Intel-Mac runtime behavior is unchanged by the drop (no RID fallback path exists).

Issue #767, opened 2026-07-24
P1: Make LSP analysis immutable, cycle-safe, observable, and E2E-tested


juanmicrosoft:
## Summary

Make LSP document/workspace analysis immutable, version-safe, cycle-safe, observable, and covered by reliable live process-level tests.

## Evidence and impact

- A concurrent dictionary contains mutable `DocumentState` instances that are reset and repopulated field-by-field.
- Handlers can observe source, tokens, AST, binding, and diagnostics from different versions.
- Stale/out-of-order versions are not rejected.
- Binder/validator exceptions are swallowed, silently producing incomplete state.
- Recursive base-member lookup has no visited set and can recurse indefinitely on inheritance cycles.
- Live hover/completion tests are skipped because the client-capability setup hangs; the harness relies on fixed delays.

## Required implementation

1. Replace mutable analysis fields with immutable, versioned `DocumentSnapshot` values.
2. Analyze off-thread with cancellation and atomically publish only if the result is still current.
3. Reject or ignore out-of-order document versions.
4. Make each handler capture exactly one snapshot for the request.
5. Convert internal analysis failures into `Calor9999` diagnostics and structured logs; never silently swallow them.
6. Add visited-type tracking and deterministic cycle diagnostics to inheritance traversal.
7. Integrate immutable workspace-index snapshots from the symbol-index work.
8. Repair the JSON-RPC test client: real capabilities, readiness signaling, bounded timeouts, deterministic teardown, and no fixed sleeps.
9. Reenable hover/completion tests and add live rename/references/diagnostics flows.

## Required regression coverage

- High-volume concurrent updates, reads, cancellation races, and out-of-order versions.
- Invariant that source, tokens, AST, binding, and diagnostics share one version.
- Injected binder/validator failures and observable diagnostics/logs.
- Direct and indirect inheritance cycles across files.
- Process-level initialize/open/change/hover/completion/definition/references/rename/shutdown.
- Repeated CI execution to detect flakes and leaked child processes.

## Success metrics / definition of done

- No handler can observe a partially updated document.
- An older version never replaces a newer snapshot.
- Every internal analysis exception is observable.
- Inheritance cycles terminate within a bounded traversal and produce deterministic diagnostics.
- All advertised core LSP capabilities have at least one enabled live E2E test.
- The live E2E suite passes 100 consecutive local repetitions and the configured CI stress repetition without a hang or leaked process.

## Dependencies

[...]

PR #921, opened 2026-08-11
Fix #765: make LSP refactoring symbol-exact

### LINKED ISSUE #765: P0: Build exact-span symbol indexing for safe LSP refactoring
## Summary

Replace LSP name-only rename/reference logic and whole-node edit ranges with exact identifier spans, stable bound symbol identities, and a closed-file workspace index.

## Evidence and impact

- Rename and references resolve a cursor to a name, then scan open ASTs for matching text.
- Shadowed locals, overloads, same-named members, and unrelated files are conflated.
- `AddNameSpan` records the containing AST node span rather than the identifier token. A rename edit can replace an entire function, statement, or expression.
- Workspace lookup scans only open documents.
- Current tests check that at least one result exists; they do not apply edits and recompile.

## Immediate containment

- Disable rename capability advertisement and reject rename requests until exact-span/symbol identity infrastructure is active.
- Keep references/definition results clearly experimental if they remain name-based during the transition.

## Required implementation

1. Define stable `SymbolId` identities for modules, types, overload signatures, members, parameters, and scoped locals.
2. Have parser/binder retain exact definition and reference identifier-token spans.
3. Build a per-document occurrence index keyed by `SymbolId`.
4. Build a project/workspace index for all `.calr` files, including closed documents, with incremental invalidation.
5. Resolve cursor positions through exact occurrence spans rather than line/name heuristics.
6. Implement rename, references, and definition from the symbol index.
7. Use versioned `documentChanges` for open documents and safe edits for closed files.
8. Remove production name-only collectors after feature parity.

- Same local name in sibling and nested scopes.
- Same method name in unrelated types.
- Overloads with identical names and different signatures.
- Fields shadowed by parameters/locals.
- Same symbol spelling across modules.
- Closed-file definitions and references.
- Exact edit range assertions for every returned edit.
- Apply edits to the workspace, then parse/bind/compile every affected file.
- Cursor positions adjacent to identifiers and on whitespace.

- Zero false-positive edits across the adversarial collision suite.
- Every rename edit range equals exactly one identifier token.
- References and definitions include closed workspace files.
- Overloads and shadowed names resolve by identity, not spelling.
- Applying all edits leaves the workspace parseable, bindable, and Roslyn-clean.
- Name-only AST collectors are absent from production handlers.

## Dependencies

Depends on stable bound symbols and exact parser/token span contracts.

juanmicrosoft:
## Summary
- index exact definition/reference token spans by stable SymbolId across open and closed workspace files
- implement definition, references, and rename exclusively through the workspace symbol index
- emit versioned document changes for open documents and safe edits for closed files
- add adversarial collision and applied-edit/rebind/Roslyn-clean gates
- enable rename by default after the safety gate

## Validation
- Release build clean
- 8,725 tests passed, 18 skipped

Closes #765


juanmicrosoft:
Adversarial review (lens: silent capability regression under a default-on switch) — one blocker found and fixed in 654fe39f.

**Finding: rename silently split declarations that span files.**

A module, and a type declared `partial` across several files, are one declaration in the language but one SymbolId *per file* in this index (`moduleId` is keyed on `GetCanonicalSourceIdentity(uri)`; each file's `TypeSymbol` has its own id). So `FindSymbolOccurrences` returned a set covering a single part, and rename applied edits to that part only.

Reproduced against this PR's own gate corpus shape:

```
definitions.calr:  §M{defs:Collision} / §CL{c001:Worker:pub:partial} / §MT{m001:Pick:pub}
open-use.calr:     §M{openUse:Collision} / §CL{c002:Worker:pub:partial} / §R §C{Pick} …
```

Renaming either `Collision` or `Worker` produced an edit touching **1 document**. Post-edit: **no Calor diagnostic** (each file still parses and binds independently) — the break appeared only after emission, as `error CS0103: The name 'Pick' does not exist in the current context`. That is the worst failure shape for a refactoring tool: it corrupts a multi-file program and the language's own analysis says the result is fine.

This matters specifically because the PR moves rename from `CALOR_LSP_EXPERIMENTAL`-gated to on by default.

**Fix.** Mark such declarations at index-build time (`ProjectSymbolOccurrence.IsSplitDeclaration`) and fail closed in `RenameHandler`, consistent with how it already refuses ambiguous, mismatched-span, and stale occurrences. A module also counts as split when another file imports its namespace, since `§U{...}` directives are not indexed as occurrences and would be left pointing at a namespace that no longer exists.

**Pins** live in `ExactSpanRefactoringGateTests`, so the dedicated CI gate job covers them: partial-type, multi-file-module, and imported-module rename must return null, and — to keep the refusal narrow — method rename across the same workspace must still apply. All three fail with the guard removed; the single-file module rename test still passes untouched.

Not fixed, worth a follow-up under #767: renaming *all* parts of a partial type or module is the correct end state, and needs identity groups in the index rather than a refusal. Filing separately.

Local: Release build clean, 11 test projects green (LSP 404 passed / 10 skipped).

Issue #943, opened 2026-08-12
[0.25 F1] Complete and validate existing ref/out call semantics


juanmicrosoft:
Parent: #1425. Depends on #1427, #1423.

## Governing 0.25 scope
Complete only the reproduced residuals in the registered ref/out design. Parsing and modifier transport already exist; retain them as regression controls rather than claiming new syntax. Reuse argument modifier AST data and preserve source semantics; do not inject Option wrappers. Cover statement/expression, named/reordered calls, overload mapping, writable/definitely-assigned operands, aliases, evaluation order and exceptions in the supported matrix. Propagate conservative mutation effects, nullable state invalidation and verifier refusal/demotion. Wire Calor/C# emitters, converter, binder, visitors, IDs, feature registry, diagnostics and relevant CLI/API/LSP paths.

### Closure
- [ ] Supported TryParse and selected dictionary TryGetValue witnesses mutate the actual caller storage and preserve return behavior.
- [ ] Unsupported declarations/discards/lvalues/generic/attribute cases reject explicitly or preserve original C# under the registered migration contract.
- [ ] Negative witnesses fail if modifiers disappear or a call becomes a copy; existing no-modifier calls remain unchanged.
- [ ] Identify an existing executable BCL witness or restore and register a named fixture; the historical BCL-06 path is currently absent and is not an acceptance prerequisite by assertion. Retain before/after provenance for the actual fixture.

## Existing behavior and success accounting
Preserve public `§A{ref|out|in}` grammar, emitter round trips and existing overload/mutation tests. Distinguish already-passing controls from new repairs in release notes. Every selected TryGetValue row must name key/value types and reference annotations supported by the frozen 0.22 contract; generic or conditional-nullability cases outside that contract must remain explicitly preserved/refused, not count as native successes. Applicable 0.24 repair blockers are added by R0 before implementation.
## Incremental delivery and decision ownership (2026-09-10)

This family owns its code, behavioral controls and family-specific website
examples. It may close when its own READY records in #1428 and #1429 pass,
without waiting for the other five families or closure of those umbrella issues.
The relevant shared test/website checks must be implemented and passing first.
READY requires the completed applicable 0.24 repairs and #1423, an exact
candidate/fixture identity, no unapproved case deletion, all expected outcomes,
and independent reviewer approval recorded by case ID and PR link.

R0 #1426 names the reviewer and obtains accepted capacity. That reviewer owns
oracle adequacy, regression classification, count-change explanations and public
[...]

PR #944, opened 2026-08-12
Pre-register the §3.1 metadata-binding entry spike

juanmicrosoft:
Registers the entry spike that gates 0.14 (roadmap §3.1) **before any binding is attempted**, so neither the denominator nor the adjudication rule can be chosen after seeing results.

## What is registered

Six BCL call-shape families, each with a `.calr` fixture: overload resolution, generic substitution, extension methods, `params` arrays, nullable annotations, `ref`/`out`.

Overload resolution leads because the tier1a postmortem blames exactly that omission — 18 overloads on `Console.WriteLine` — for the last broken estimate in this area.

Adjudication is a **fixed GREEN/AMBER/RED table with no "try harder and re-measure" row**, and resolving to the *wrong* overload counts as **wrong**, not as resolved. That distinction is the point: a plausible wrong answer is what this area produces, and unresolved is honest where wrong is not.

## Writing the fixtures found something, before measuring

**Calor has no call-site syntax for `ref`/`out` arguments** (#943). It can *declare* an `out` parameter — the converter emits `(str:input, i32:result:out)` — but it cannot *pass* one. `§O` is a statement-level output declaration, and every call-construction site in the parser passes `argumentModifiers: null`. `Int32.TryParse(text, out value)` is unwritable.

That is a **language gap, not a binding gap**, and my first draft of the adjudication table would have scored it as a binder failure — potentially pushing a working mechanism to RED over a call nobody can write. The table now separates *not-expressible* from *unresolved*, and the counts run over the five expressible families.

Catching that while still pre-measurement is precisely what the registration discipline is for. Had I measured first, the correction would have looked like moving the goalposts.

## BCL-06 is kept, not deleted

Registered on the binder ratchet's allowed-parse-failure list against #943, so the gap stays visible and is re-tested automatically once the syntax lands. Deleting it would have quietly shrunk the spike's denominator to the shapes that happen to work — which is the failure mode the whole document exists to prevent.

## Baselines

Binder ratchet 506 → 511 parsed, 9 → 10 registered parse failures, `IncompleteCount` still 0. Formatter corpus 877 → 883 tracked, 642 → 646 losslessly formatted, with BCL-05 registered as a semantic fallback and BCL-06 as a parse failure.

Release build clean; all test projects green; test-quality, dogfood, and Calor-first guards pass.

**§5 (Results) is deliberately empty.** Measurement is the next PR.

### DIFF
--- a/bench/phase0-agent-native/binder-incomplete-baseline.json
+++ b/bench/phase0-agent-native/binder-incomplete-baseline.json
@@ -1,8 +1,8 @@
 {
   "IncompleteCount": 0,
-  "ParsedFiles": 506,
-  "ParseFailures": 9,
-  "ExpressionsBound": 4702,
+  "ParsedFiles": 511,
+  "ParseFailures": 10,
+  "ExpressionsBound": 4721,
   "Scope": "both F-2 legs; conversion leg skips when corpus submodules are absent",
   "Conversion": {
     "Incomplete": 0,
--- a/docs/plans/spike-3-1-bcl-binding-registration.md
+++ b/docs/plans/spike-3-1-bcl-binding-registration.md
@@ -1,0 +1,112 @@
+# §3.1 entry spike — pre-registration
+
+**Status: registered, not yet measured.** This document exists to be committed
+*before* any binding is attempted, so the denominator and the adjudication rule
+cannot be chosen after seeing results. Nothing in §2 or §3 may be edited once
+measurement begins; §5 is filled in afterwards.
+
+## 1. Why this spike gates 0.14
+
+The roadmap (§3.1) names metadata-backed .NET binding as the component that can
+sink 0.14 — not the flow checker, which has TIER1A. It is a slice of Roslyn's
+binder: exact receiver type, overload resolution, parameter and return types,
+generic substitution, nullable annotations.
+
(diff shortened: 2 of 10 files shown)


juanmicrosoft:
Closing as superseded, per roadmap Draft v4 §4.3 (M1: "register-then-merge has a precedent to repair first").

This PR was the §3.1 pre-registration, but the spike shipped through a different registration path while it sat open: the scoping doc merged as PR #987 (`docs/plans/v0.14-metadata-binding-scoping.md`), the BCL call-shape fixtures as PR #1019 (S3, 15 shapes), and the resolved-fraction ledger + anti-tautology pin as PR #1022 (S5). The binder-incomplete baseline and ratchet test this PR also touched have since moved on main.

Recorded discrepancy, so it is not laundered: the registration that governed the spike (#987) merged 2026-08-18, the same day as S1 (#988) — register-then-merge held by a day, not by the review-cycle margin this PR intended. The 0.15 PP registration (A-1.10) is required to merge before any effect-row implementation (§4.3 (i)), and the A-1.10 PR extends `experiment-registry-tamper-check.yml` to the annex so the ordering is mechanically guarded.

Nothing in this branch is lost: the fixtures and ledger exist on main under their shipped names.

Issue #965, opened 2026-08-13
Release blocked: Calor.Performance.Tests reproducibly kills the CI runner (exit 143), and the perf suite runs ONLY on the release path


juanmicrosoft:
v0.13.2 is tagged (`59ebda42`) and **cannot publish**. Two jobs in `publish-nuget.yml` die reproducibly, both while running `Calor.Performance.Tests`.

## Evidence

Run 31651714822 (release event for `v0.13.2`), original attempt and re-run:

| job | attempt 1 | attempt 2 |
|---|---|---|
| `test (performance)` | exit 143 after 3 min | exit 143 after 2 min |
| `release-quality` | 40-min timeout during perf gate | exit 143 after 10 min, during perf gate |
| everything else | 17 success | 17 success |

`##[error]The runner has received a shutdown signal` + `Process completed with exit code 143` (SIGTERM). **The tests that ran were all passing** when the runner died.

Both jobs die at the same place. `test (performance)` runs the suite directly; `release-quality` runs `run_mutation_gate.py` (passes — all 7 mutants killed), then `run_performance_gate.py`, which runs the same suite four times.

In both `test (performance)` attempts the last test to complete was `CFGConstruction_SmallFunction_Under10ms`, immediately before the `Memory_*` tests in local execution order.

## It is not the tests being slow or leaky

Locally on the tagged commit (`59ebda42`, includes #960 and #961):

```
Total tests: 21   Passed: 21   Total time: 19.6 s     (baseline: expectedTestCount 21, maximumMedian 21.0s)
Medium module memory:  2.66 MB   (ceiling 50 MB)
Small module memory:  -0.08 MB   (ceiling 10 MB)
Large module memory:  18.19 MB   (ceiling 100 MB)
```

So the suite is within every budget it declares, and the measured memory deltas are nowhere near the thresholds. Whatever terminates the runner is not visible in the assertions.

## The structural problem this exposed

**The performance suite runs only on the release path.**

- `test.yml`'s PR matrix has **no `performance` leg** — check the reported contexts on any recent PR.
- `quality-ratchets` runs `run_performance_gate.py --self-test` only, never the real gate.
- `publish-nuget.yml` runs both the real `test (performance)` leg and the real perf gate.

[...]

Issue #949, opened 2026-08-12
Branch protection requires a status check that no longer exists — 'enforcement-tests' enforces nothing and blocks every PR


juanmicrosoft:
Found while checking why `gh pr merge --admin` was needed on the v0.13.1 release PR.

## The mismatch

`main`'s branch protection requires four status checks:

```
test, enforcement-tests, calor-first-guard, d-s1.5-fixtures
```

Three of those still report. **`enforcement-tests` does not.** The CI restructure in #938 moved the enforcement suite into the `tests` matrix, where it now reports as **`tests (enforcement)`**.

Full check list on a current PR contains `tests (enforcement)`; there is no `enforcement-tests`.

## Two consequences, and the second is the bad one

**1. Every PR is permanently `BLOCKED`.** GitHub treats a required check that never reports as perpetually pending, so `mergeStateStatus` is `BLOCKED` on every PR regardless of results. Combined with `required_approving_review_count: 1`, `--admin` becomes the routine way to merge — which is how the first consequence hides the second.

**2. A quarter of the required-check protection enforces nothing.** A check that never runs cannot fail. Enforcement-suite regressions are no longer gated by branch protection; they are caught only by whoever reads the check list. The protection *looks* configured — the context is listed in the API response — and is inert.

This landed in the PR whose stated purpose was making CI and release quality gates truthful (#938), which is precisely why it is worth fixing rather than shrugging at: the rename was correct, the protection config simply was not carried with it.

## Fix

Update the required contexts to the current job names, e.g.:

```
test, tests (enforcement), calor-first-guard, d-s1.5-fixtures
```

Worth reviewing at the same time whether the required set should be broader now that the matrix exists — `tests (compiler)`, `tests (verification)` and `quality-ratchets` are all release-critical per `eng/test-manifest.json`, yet none of them gate merging.

## Related

- #948 — `roundtrip-verification` flake. Note it is **not** a required check, so it never gated merges either.
- #897 — `McpServerTests` flake, which surfaces through `quality-ratchets`, also not required.

juanmicrosoft:
Verified on 2026-08-27 as part of the v0.16 kickoff sweep (https://github.com/juanmicrosoft/calor/pull/1122). `gh api repos/juanmicrosoft/calor/branches/main/protection --jq '.required_status_checks.contexts'` now reads:

```
["test","calor-first-guard","d-s1.5-fixtures","tests (compiler)","tests (enforcement)"]
```

[...]

PR #976, opened 2026-08-14
docs: performance gate strategy — fix the calibration, not the metric (#965)

juanmicrosoft:
**Revision 2.** Revision 1 proposed deleting the aggregate performance ratchet. Four independent reviews refuted its premise, its evidence, and its remedy. Rewritten around what the measurements support. Still docs-only.

## The premise was false

All four failed v0.13.2 publish runs were read log-by-log. **Not one printed `ERROR: performance median exceeded the ratcheted maximum`** — every blocking failure was `exit 143` / runner shutdown or a test-host crash. And `test (performance)`, one of the two dying jobs, **never invokes the gate script**.

Retiring the ratchet would have unblocked nothing. The real blocker is #965, still open.

## The metric was sound; the calibration was wrong

| | suite | startup | compute | startup share |
|---|---|---|---|---|
| dev | 19.871s | 0.764s | 19.107s | **3.8%** |
| CI | 21.348s | 1.567s | 19.781s | **7.3%** |

It is **92–96% compute**. Revision 1's "wall-clock is mostly startup" conflated the *level* with the dev-vs-CI *difference*. The aggregate is in fact the most stable signal in the suite — **CV ~1.6%** across 18 CI samples on 6 runners — and the 21.0s ceiling simply sat *inside* the 20.352–21.439s distribution.

## The proposed replacement was worse

Measured flake rate with **no code change**, budgets at k × CI median:

| k | P(red per run) |
|---|---|
| 2× | **12.2%** |
| 2× calibrated on a dev machine | **93%** |
| 3× | 0% of 24 samples |

Plus: class ordering alone swings individual tests **1.78–1.90×** (whichever class runs first pays JIT); sub-100ms tests assert on *integer* milliseconds; **three memory tests measure ~0.00 MB and cannot fail** (`MeasureMemory` collects after the module is unrooted); and `CFGConstruction_SmallFunction_Under10ms` — revision 1's exhibit for tightness — has **~250× headroom, not 2×**.

## Executing it would have red-lit a required check

8 files, not 4. `test.yml:249` runs the script on **every PR**, and `check_test_quality.py` — inside the **required** `calor-first-guard` check — satisfies the manifest→workflow link *only* by reading the script being deleted. That is #949's failure mode: a required check failing for reasons unrelated to the PR.

## Revision 2

1. **Unblock by fixing calibration** — ceiling 21.0 → 24.0s with the evidence recorded, per the #942 precedent (which set 20.0 → 21.0 the same way). One line.
2. **Root-cause #965** — the actual blocker. Carries the facts that matter: kills at 24s/26s/52s/79s/26m, `always()` steps *skipped* (⇒ cancellation), healthy host pre-gate, and that exit 143 is classically OOM while Calor links **Z3 native** — invisible to `GC.GetTotalMemory` and so to all 21 budgets.
3. **Repair the substrate** — vacuous memory tests, class ordering/JIT, integer-ms assertions, `Scalability_Linear*` which permits O(n²).
4. **Add deterministic counters** — measured here: token count and CFG-block count **identical across all 25 measurements**, allocated bytes **1.20% spread**, versus **123%** for wall-clock on the same operation.

## §7 keeps the retraction

Every refuted claim is listed with what refuted it, including that my dev measurement (19.871s) **does not reproduce** — 16.674s same machine, same branch, same day, which collapses revision 1's headline 3.5% result to ~25%.

The pattern worth naming: I matched "this needs three layers of correction, so the metric must be wrong" onto a case where the metric was sound and my calibration was wrong — and nearly deleted the most stable signal in the suite to fix a release it was not blocking.

## Decisions — §8

1. Ceiling raise to 24.0s, or the compute-subtracted gate at 22.0s (already built, 3/3 green)?
2. Do substrate repairs block the counters? (recommend no — counters are order- and JIT-independent)
3. Aggregate stays release-blocking, or nightly-only?

### DIFF
--- a/docs/plans/performance-gate-strategy.md
+++ b/docs/plans/performance-gate-strategy.md
@@ -1,0 +1,190 @@
+# Performance gate — strategy change
+
+Proposes retiring the aggregate wall-clock ratchet added by #938 and returning
+the performance gate to per-test budgets, then strengthening those budgets.
+
+Written because the aggregate gate has blocked the v0.13.2 release, has never
+passed reliably in CI, and required three successive layers of statistical
+correction — each of which either failed or made things worse. That pattern is
+the argument, not the individual failures.
+
+This document fixes the strategy and the evidence behind it. It does not
+re-tune every per-test budget; that lands with the follow-up in §6.
+
+## 1. What the gate is for
+
+One question: **did a change make the Calor compiler slower?**
+
+A useful answer has to be *actionable* (name what regressed), *portable* (mean
+the same on a laptop and a shared runner), and *sensitive* (fire on a real
+regression, stay quiet otherwise). The current arrangement fails all three.
+
+## 2. What exists today
+
+Two gates, stacked, measuring overlapping things:
+
+**(a) Per-test budgets — 21 tests, each asserting its own.** The budget is in
+the test name: `Parsing_MediumModule_Under500ms`,
+`CFGConstruction_SmallFunction_Under10ms`, `Memory_LargeModule_Under100MB`. A
+failure names the operation that regressed.
+
+**(b) An aggregate wall-clock ratchet** (`scripts/run_performance_gate.py` +
+`eng/performance-baselines.json`), added by #938. Runs the suite 4× and caps the
+median at `maximumMedian` (21.0s).
+
+(b) is the subject of this document. (a) stays either way.
+
+## 3. Why the aggregate is the wrong measurement
+
+Not "badly tuned" — measuring the wrong quantity.
+
+### 3.1 It is dominated by the term we do not care about
+
+The suite's wall-clock is mostly process start, assembly load, JIT and test-host
+overhead. Measured 2026-08-14, same commit:
+
+| | suite wall-clock | startup (calibration) | compute |
+|---|---|---|---|
+| dev machine | 19.871s | 0.764s | 19.107s |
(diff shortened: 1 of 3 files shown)


juanmicrosoft:
Closing in favour of the v0.16 kickoff sweep PR, https://github.com/juanmicrosoft/calor/pull/1122, with this PR's parts dispositioned separately:

- **`docs/plans/performance-gate-strategy.md`** — carried into the tree as written (with a dated status note on top, credited to this PR). `eng/performance-baselines.json`'s `noisePolicy` has cited it by path since #978, so it was a dangling reference until now.
- **`eng/performance-baselines.json`** — already on main via #978 (phase 1, ceiling 21.0 → 24.0s). Not carried.
- **`scripts/run_performance_gate.py`** (the compute-subtracted gate) — the road not taken once #978 raised the ceiling. Not carried.
- **Phases 3–4** (vacuous memory tests, class ordering/JIT, integer-ms assertions, `Scalability_Linear*`; deterministic counters) — re-filed as #1115 so they are tracked as work rather than as prose.

Status of phase 2 / #965 as of today: the v0.15.0 release run (33096892494) was green on `test (performance)` and `release-quality`; no exit 143 since the 0.13.2 attempts. Roadmap-v0.16 §7 makes #965 release-blocking for 0.16.0 only if it recurs.

Issue #784, opened 2026-07-24
P0: Make taint analysis report direct injection and track symbols through CFG


juanmicrosoft:
## Summary

Reimplement taint analysis as a symbol-keyed CFG may-taint analysis so direct source-to-sink flows are always reported and sanitization/reassignment behavior is sound.

## Evidence and confirmed reproductions

- Default CLI configuration suppresses findings with fewer than two taint hops.
- A direct `user_input -> db.execute` flow is therefore absent unless `--all-findings` is used.
- State is a name-keyed linear dictionary across branches.
- Some call-source assignments, including `Console.ReadLine`, are missed.
- Clean reassignment does not reliably clear taint.
- Sanitizers are detected by substring, so a method named `desanitize` can suppress a finding.

## Required implementation

1. Report direct source-to-sink flows by default; hop count may rank findings but must not suppress canonical injection.
2. Implement a forward CFG may-taint lattice keyed by stable symbols and access paths.
3. Model branch joins, loops, aliases, fields, collections, returns, and interprocedural summaries.
4. Apply strong updates for safe reassignment where aliasing permits.
5. Resolve source/sink/sanitizer behavior by symbol/signature/manifest, not substrings.
6. Define sanitizer preconditions and output taint precisely.
7. Track provenance/path evidence for diagnostics.
8. Mark unknown external behavior conservatively under strict analysis.

## Required regression coverage

- Direct one-hop source-to-sink flows.
- `Console.ReadLine` and other call-return sources.
- Safe overwrite, branch-only taint, loop propagation, aliases, fields, and collection elements.
- Real sanitizer versus `desanitize`/similar-name methods.
- Interprocedural return/parameter propagation.
- Multiple sources and sinks with path evidence.
- False-positive controls for constants and validated values.

## Success metrics / definition of done

- The canonical direct injection reports under default settings.
- All configured source/sink/sanitizer matches use resolved identities.
- Safe strong-update fixtures stop reporting while unsafe branch/alias fixtures continue reporting.
- Every finding includes a reproducible source-to-sink path.
- Precision/recall targets are committed for the adversarial security corpus and enforced in CI.
- `--all-findings` changes verbosity/ranking, not whether direct exploitable flows exist.

## Dependencies

Depends on typed bound IR, stable symbols, and corrected CFG/dataflow.

Issue #782, opened 2026-07-24
P0: Enforce refinement and proof-obligation guards using program state


juanmicrosoft:
## Summary

Wire refinement and proof-obligation policy into real runtime guards, and make obligation proofs account for program state, control flow, assignments, and nesting.

## Evidence and impact

- Public refined parameters are classified as boundary obligations and policy maps them to `AlwaysGuard`, but `RequiresGuard` has no production consumer.
- Refinement/indexed types are erased during emission.
- Return/subtype obligation kinds exist but are not generated.
- The solver declares parameters and size variables but does not symbolically execute the function body.
- Assignment kill/state updates are incomplete.
- Loop facts use an exclusive upper bound while generated loops are inclusive.
- Proof generation sees only some direct body nodes; nested proofs can degrade to TODO comments without a guard.

## Required implementation

1. Inject policy-driven guards at parameter entry, return, assignment/coercion, collection/index access, and explicit proof sites.
2. Generate all defined obligation kinds, including refinement return and subtype transitions.
3. Define guard behavior for proven, failed, refuted, timeout, unknown, unsupported, and internal error outcomes.
4. Traverse the full AST/CFG so nested obligations cannot be skipped.
5. Replace fact-only body reasoning with CFG/SSA symbolic state or a conservative equivalent.
6. Model assignments, compound assignments, aliases, branches, loops, catches, and mutations.
7. Correct loop facts to match direction, inclusivity, and actual emitted semantics.
8. Preserve a runtime guard whenever proof state is incomplete.

## Required regression coverage

- Public/private refined parameters and returns.
- Refinement assignments and subtyping.
- Indexed bounds at reads and writes.
- Nested obligations under every statement container.
- Reassignment invalidating a prior fact.
- Inclusive positive/negative loops and final iteration.
- Timeout, unsupported, and solver failure retaining guards.
- Runtime tests that intentionally violate each boundary.

## Success metrics / definition of done

- Every generated obligation has one explicit policy decision and enforcement site.
- `AlwaysGuard` and `WarnAndGuard` result in executable guards.
- Violating every public refinement fixture throws at runtime.
- Nested obligations are counted and enforced identically to top-level obligations.
- No proof relies on loop facts that disagree with emitted iteration semantics.
- Unknown/timeout/unsupported/error outcomes never remove protection.

## Dependencies

Depends on structural CFG/bound IR. Verification containment remains active until complete.

Issue #781, opened 2026-07-24
P1: Correct contract inheritance and preserve modules in contract transforms


juanmicrosoft:
## Summary

Make contract inheritance mathematically correct across clause conjunctions, overloads, and multiple interfaces, and make contract transforms preserve every unrelated module construct.

## Evidence and impact

- Contract inheritance checks each interface clause against any implementation clause rather than proving implication between conjunctions.
- An implementation may add preconditions when the interface has none.
- Inherited contracts are keyed only by class and method name, so overloads and multiple interfaces overwrite one another.
- Emitter lookup is similarly name-based.
- Contract simplification reconstructs a module through an older constructor/overload and can drop enum extensions, C# interop blocks, refinement/indexed types, and preprocessor blocks whenever any contract changes.

## Required implementation

1. Formalize Liskov contract rules:
   - implementation preconditions must be no stronger;
   - implementation postconditions must be no weaker.
2. Prove implications over the conjunction of all relevant clauses.
3. Treat an unconstrained base/interface as `true`; adding implementation requirements must fail.
4. Key inherited contracts by resolved method identity/signature.
5. Combine contracts from all implemented interfaces/base members deterministically.
6. Diagnose incompatible inherited contract sets.
7. Replace positional module reconstruction with immutable `with`/copy APIs that preserve every field.
8. Add structural invariants for every AST transformation pass.

## Required regression coverage

- Multiple pre/post clauses whose conjunction differs from pairwise matching.
- Base with no preconditions and implementation with one.
- Overloads with the same method name.
- Multiple interfaces with compatible and incompatible contracts.
- Generic methods and explicit interface implementations.
- Module containing every optional field while a contract is simplified.
- No-op transform preserving a semantic hash of all noncontract nodes.

## Success metrics / definition of done

- Contract inheritance results match conjunction-based implication for the complete matrix.
- Additional implementation preconditions against an unconstrained base are rejected.
- Overloads and multiple interfaces never overwrite each other.
- A contract-only transform changes zero unrelated AST nodes/collections.
- Reflection/constructor tests fail when a new `ModuleNode` field is not preserved by a transform.

## Dependencies

Uses typed method/symbol identity and exact verifier semantics. Structural preservation can land independently.

juanmicrosoft:
[...]

PR #981, opened 2026-08-15
docs: describe the ten changes sitting unreleased on main

juanmicrosoft:
`[Unreleased]` was empty while ten changes sat on top of the v0.13.2 tag. This writes them down.

## Where the ten sit against the roadmap

Set against §5.2 of `docs/plans/roadmap-v0.13-v0.15.md`:

- **One lands on schedule** — #764, a 0.13 MUST.
- **Six close issues scheduled for a later version** — #782 (0.14), #767 (0.13.x/0.14), and #763, #784, #785, #786 (0.15). The analysis and emission layers are ahead of the staircase.
- **One, #781, was deferred** to 0.15's §4.1 design-doc scope decision rather than scheduled.
- **Two, #766 and #768, were not scheduled at all** — §5.2 lists them in the row demoted to demand-driven work, re-opened only by a named failing fixture. They landed anyway.

(An earlier revision of this PR said "eight close issues scheduled for 0.14 or 0.15" and named #766/#768 among them. That was wrong in both directions — those two were demoted, not scheduled, and it omitted #767, which was. Corrected against the table.)

## The entries that matter

Several are **behaviour changes that newly reject or newly diagnose code v0.13.2 accepted**:

- effect enforcement fails closed on operations it cannot resolve, instead of assuming purity from a name match (#968)
- taint reports direct-injection findings by default — existing code gets new diagnostics with no source edit (#969)
- inherited contracts incompatible with an override are now diagnosed (#966)
- postconditions on iterators are rejected, and opaque raw-C# returns fail closed (#967)
- generated C# is standalone, with implicit usings disabled (#977)
- bug-pattern findings are re-partitioned into verified / heuristic hint / incomplete analysis (#970)

Each is marked as a behaviour change rather than written up as an improvement.

## Provenance

Entries are derived from the merged pull requests named in each line — a record of what those PRs claim to have done, not an independent re-verification of each.

The one `Fixed` entry says plainly that **#965 remains open and is not root-caused**: raising the perf ceiling from 21.0s to 24.0s (#978) fixed a calibration error — the ceiling had been derived from a dev machine and sat inside the CI distribution — but the exit-143 runner terminations behind four failed publishes are still unexplained.

## What this does not do

The website changelog mirror is left alone; it is written at the release cut by the `create-release` skill, and publishing unreleased content to the public site would misdescribe what is installable.

It also does not decide **0.13.3 vs. folding into 0.14**. `[Unreleased]` is correct either way. My recommendation is a 0.13.3 cut: these are breaking-ish semantics changes, and bundling them into 0.14's notes would make it impossible to separate "the null-safety break" from "the effects and analyzer break" for anyone reading why their program stopped compiling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,90 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+Ten changes have landed on `main` since the v0.13.2 tag. Set against §5.2 of
+`docs/plans/roadmap-v0.13-v0.15.md`, they sit like this:
+
+- **One lands on schedule**: #764 (structural postcondition lowering) is a 0.13 MUST.
+- **Six close issues scheduled for a later version**: #782 (0.14), #767 (0.13.x/0.14), and #763,
+  #784, #785, #786 (0.15) — so the analysis and emission layers are ahead of the staircase.
+- **One, #781, was deferred** to 0.15's §4.1 design-doc scope decision rather than scheduled.
+- **Two, #766 and #768, were not scheduled at all**: §5.2 lists them in the row demoted to
+  demand-driven work, re-opened only by a named failing fixture. They landed anyway.
+
+Several are behaviour changes that can newly *reject* or newly *diagnose* code that v0.13.2
+accepted; those are marked below. Entries are derived from the merged pull requests named in each
+line.
+
+### Changed
+
+- **Effect enforcement is symbol-resolved, exhaustive, and fail-closed (#968, closes #785).** Calls,
+  constructors, accessors, overloads, delegates and inherited members resolve by symbol and
+  signature instead of bare-name purity heuristics, and allocation, initialization, disposal, event,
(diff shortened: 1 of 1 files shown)


juanmicrosoft:
Closing in favour of the v0.16 kickoff sweep PR, https://github.com/juanmicrosoft/calor/pull/1122, which carries this write-up rather than superseding it.

Checked before closing: none of the eleven PRs this describes (#968, #970, #969, #972, #966, #967, #964, #975, #977, #980, #978/#974) appear in the `[0.14.0]` notes by number or by keyword ("fail-closed", "direct-injection", "implicit usings", "24.0s"), yet all of them merged 2026-08-13 to 2026-08-15 and shipped in 0.14.0 (2026-08-23). So `[Unreleased]` is no longer the right home; the sweep PR adds them as a dated "recorded late" subsection under `[0.14.0]`, rewritten in plainer language per the CHANGELOG readability rule, crediting this PR. The "0.13.3 vs fold into 0.14" question this PR left open was answered by events (folded).

Issue #790, opened 2026-07-24
P0: Turn CI and regression tests into truthful release gates


juanmicrosoft:
## Summary

Turn CI and regression tests into truthful release gates: make round-trip verification blocking, run every test project, collect meaningful coverage, eliminate vacuous assertions, and reenable skipped live functionality tests.

## Evidence and impact

- External round-trip verification is `continue-on-error`.
- Performance tests are not in the solution or CI.
- Coverage collectors are referenced but CI collects no coverage and enforces no threshold.
- There is no mutation-testing gate.
- ID validation masks command failure with `|| echo` and checks `examples` rather than the actual `samples` directory.
- Multiple obsolete skipped tests remain; live LSP hover/completion tests are skipped because the harness hangs.
- Behavioral tests use assertions such as `|| true`, `Count >= 0`, or only `Assert.NotNull`.
- Several suites run redundantly while important projects/gates remain absent.

## Required implementation

1. Make round-trip verification blocking after establishing an explicit baseline.
2. Add every maintained test project, including performance, to a CI-owned test manifest/solution.
3. Collect line/branch coverage and establish ratcheting per-component thresholds.
4. Add mutation testing for verifier, emitter, binder, migration, dataflow, taint, and effect safety logic.
5. Remove failure-masking shell constructs and validate `samples/`.
6. Delete or rewrite obsolete skipped tests; fix and reenable live LSP tests.
7. Add a test-quality analyzer/lint rule banning known vacuous assertion patterns.
8. Consolidate redundant jobs while retaining platform/RID coverage.
9. Publish test, coverage, mutation, performance, migration, and flake reports as artifacts.
10. Define which gates block a release and enforce them in the release workflow.

## Required regression coverage

- A deliberately failing ID/sample validation command must fail CI.
- A deliberately type-invalid round-trip fixture must fail CI.
- A killed mutant in each safety-critical component.
- Performance regression fixture proving threshold enforcement.
- Skipped-test inventory check.
- CI test manifest check proving no test project is omitted.

## Success metrics / definition of done

- 100% of maintained test projects run in CI.
- Zero release-critical jobs use `continue-on-error` or failure-masking `||` fallbacks.
- Zero behavioral tests contain the banned vacuous assertion patterns.
- Coverage and mutation scores are published and cannot decrease below ratcheted baselines.
- Performance thresholds are enforced with documented noise controls.
- Core LSP features have enabled live E2E coverage.
- The release workflow depends on every declared release gate.

[...]

PR #982, opened 2026-08-15
ci: close roadmap gate 7's CLI leg, and fix what building it found

juanmicrosoft:
Stacked on #981 (base retargets to `main` when that merges).

Roadmap §2.5 gate 7 was the one release gate still open at 0.13.x. Its Sdk half has been in CI as `sdk-package-consumer` since v0.11. Its CLI half — the `calor` global tool — was only ever a **manual post-publish check**, which is exactly the kind of gate the #790 rewrite was supposed to eliminate.

## The bar is a solver verdict, not exit 0

The state gate 7 was written against is *"installs successfully, silently loses verification"*. A consumer with no working Z3 still installs, still exits 0, and reports every contract `Skipped`. So `cli-tool-consumer` packs `calor`, installs it from a local feed with a cold package cache, and requires **Proven > 0, Skipped == 0, no Calor0710** — on gate 7's frozen RIDs (win-x64, linux-x64, osx-arm64). It also asserts the package carries the runner's native and *does not* carry an osx-x64 one, so the #916 drop stays a tested fact.

## The amendment, as a test instead of a promise

The 2026-08-11 amendment (#916 review F3) requires that a clean Intel-mac consumer get **documented degradation**: loud, no crash, no silent pass. There is no Intel-mac runner to depend on, so the script reproduces the condition exactly — remove the Z3 native for the runner's *own* RID from the installed tool, which is the state an osx-x64 consumer is in — and asserts all three properties.

## That oracle failed on first run, against shipped 0.13.2

`Calor0710` is `info` severity, and the verify text report prints only errors and warnings. So with the native removed:

```
File: proven-contracts.calr
  Proven:      0
  Skipped:     14
  Proven Rate: 0.0%
```

Exit 0. No mention of Z3 anywhere in the output. `--verbose` did not surface it either — the diagnostic existed **only** under `--format json`.

That is gate 7's own failure mode, live in the product, found by building the gate. "No silent pass" held — nothing was reported `Proven` — but "loud" did not, on the surface a human or agent actually reads.

The text report now names the reason above the counts it explains, and repeats it in the overall summary (on a multi-file run the per-file notice scrolls away, and `Proven Rate: 0.0%` alone does not distinguish *nothing proved* from *nothing attempted*). **Severity, the JSON envelope, and the exit code are unchanged** — this is a reporting fix, not a diagnostic-semantics change.

## Discrimination and its limit

The pin discriminates in the strongest available order: it was written first, **failed against the unfixed compiler**, and passes after. The transcript of the failing run is in the commit message.

It is an end-to-end shell gate rather than an xUnit test, deliberately. Forcing Z3-unavailable in-process means adding a mutable seam to `Z3ContextFactory`'s static availability cache, which every test in the assembly shares — this repo has already been bitten once by cross-test contamination through exactly that kind of shared static. The tradeoff is stated rather than hidden: **the loudness behaviour is pinned by CI, not by the unit suite.**

## Verification

- both legs pass locally on osx-arm64 (`Proven 14, Skipped 0`; then `Calor0710` raised, nothing Proven, exit 0, no crash)
- `Calor.Verification.Tests` 393/393
- Release build, 0 warnings
- `check_test_quality.py` passes; `shellcheck` clean; workflow YAML parses, job registered, matrix correct

## Not done here

Branch protection is unchanged. Gate 7 is a release gate, so `cli-tool-consumer (…)` arguably belongs in the required contexts — but #949 is the open record of what happens when required contexts and job names drift, and changing protection is the maintainer's call.

### DIFF
--- a/.github/scripts/test-cli-tool-consumer.sh
+++ b/.github/scripts/test-cli-tool-consumer.sh
@@ -1,0 +1,202 @@
+#!/usr/bin/env bash
+# Roadmap §2.5 gate 7 — clean-consumer install, CLI leg (#790).
+#
+# The Sdk leg of gate 7 is `sdk-package-consumer` / test-sdk-package.sh. This is
+# its counterpart for the `calor` global tool, which until now was only ever
+# verified by hand after a publish.
+#
+# The bar is deliberately NOT exit-0 install. The pre-closure osx-x64 state was
+# precisely "installs successfully, silently loses verification" — a consumer
+# whose Z3 native is missing still installs, still exits 0 on `calor verify`,
+# and reports every contract Skipped. So this script requires a *solver verdict*:
+# contracts Proven, none Skipped, no Calor0710.
+#
+# It then runs the gate's 2026-08-11 amendment (#916 review F3) as a test rather
+# than a promise. #916 dropped the mislabeled osx-x64 RID from the package, and
+# the roadmap requires that an Intel-mac consumer get *documented degradation* —
+# a loud Calor0710, no crash, no silent pass. GitHub does not offer an Intel-mac
+# runner we can rely on, so the condition is reproduced exactly on whatever
+# runner we have: remove the native for the runner's own RID from the installed
+# tool, which is the same state an osx-x64 consumer is in, and assert the
+# documented degradation.
+#
+# Env:
+#   CALOR_CLI_VERSION   override the version to pack/install (default: Directory.Build.props)
+#   CALOR_EXPECTED_RID  require the runner RID to match (CI pins this per matrix leg)
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+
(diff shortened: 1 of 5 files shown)


juanmicrosoft:
Carried into the v0.16 kickoff sweep PR, https://github.com/juanmicrosoft/calor/pull/1122 — cherry-picked as `9161e244` (the `cli-tool-consumer` job on gate 7's three frozen RIDs, `.github/scripts/test-cli-tool-consumer.sh`, the `VerifyCommand` "NOT VERIFIED: Z3 unavailable" text-report fix, and the gate-7 status paragraph in `roadmap-v0.13-v0.15.md`). The only conflict was this PR's CHANGELOG hunk against `[Unreleased]`; it was replaced by two plain-language lines. The script was run locally on osx-arm64 against 0.15.0: pack → cold-feed install → `Proven 14, Skipped 0` → native removed → `!! NOT VERIFIED … Calor0710`, nothing Proven, exit 0, no crash — PASS.

One correction for the record: roadmap-v0.16 §0.1/§6 call this PR "gate 3's CLI-process leg". It is §2.5 **gate 7**'s CLI leg (as its own title and R:1022 say); gate 3's CLI-process leg over the edit-script corpus is still unbuilt and is tracked as #1116. Closing in favour of the sweep PR.

